PR #24, opened 2022-06-18
Devcontainer & CMake support (develop branch)

nedsociety:
Substitutes #21 which is based on master branch.

---

This is a long-term draft PR for making the dev process more accessible. When it's complete only vscode and docker will be required for building and debugging Hyperspace. See BUILDING.DEVCONTAINER.md in the PR for the build instruction.

TODO:

* [x] Fix crash for `FTLGame.exe > NUL` on Windows Release build
* [x] Setup image caching @ ghcr.io
* [ ] A better configuration for calling generateVersion.sh ([see this](https://stackoverflow.com/questions/61793742/cmake-regenerate-source-file-if-target-gets-rebuilt))
* [x] A shell script to automate builds? (need to sort out the use cases)
* [ ] Github workflow for CI/CD
* [ ] VSCode: Support IntelliSense ([this issue](https://github.com/microsoft/vscode-cmake-tools/discussions/2740) might be blocking the correct setup for Windows build)
* [ ] VSCode: Support debugging with CodeLLDB extension


Nasa62:
> ### Current status
> Upgraded clang from 10 to 14 with `lld` as linker. It seems to remove most of the scary linker warnings [as expected](https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/21#issuecomment-1154802499).
> 
> Currently all the Release versions on Windows x86, Linux x86 and amd64 build successfully heavy_check_mark. The Windows version passes the launch test as well. However the Debug builds fail x on all platforms by crashing the compiler ([llvm/llvm-project#56126](https://github.com/llvm/llvm-project/issues/56126)). Note that clang-10 also failed to build Debug versions for other reasons (no ICE on compiler but fails at the linking stage).

Fun that we've found an actual compiler bug

> Here's the build outputs if you'd like to test them: [compiled.zip](https://github.com/FTL-Hyperspace/FTL-Hyperspace/files/8941504/compiled.zip)

The linux files here are drastically larger than my builds (probably because of static deps).
It's also not clear if you built these for 1.6.12 or 1.6.13 as there is a struct difference between the two and there needs to be the two separate linux binaries for them.

Either way, x86 build just segfaults.

Amd64 build doesn't work at all and can't find any signatures and also doesn't trigger any of the message prompts from libsdl2, maybe because it's statically linked?


The logs make zero sense from the AMD64 build, the build just seems completely borked because it seems to execute hyperspace multiple times? Like there's multiple segscans but that shouldn't happen at all.

```
➜  data LD_PRELOAD=./Hyperspace.amd64.so ./FTL.amd64
ELF Relocation addr: 0x0000000000000000
ELF Loaded at 0x0000000000400000, 0x0000000000400000
SEG SIZE: 0x0000000000200000
SEG SIZE: 0x0000000000200000
SEG SIZE: 0x0000000000200000
Initializing Hyperspace...
Starting Lua
Initializing Lua
Hyperspace SWIG Lua loaded
Lua initialized!
Lua Inited
Hyperspace initialized
Version: 1.6.12
Loading settings
Initializing Crash Catcher...
Starting up
Loading text
//////////////////// Skipped all the doubly defined dictionary stuff that doesn't matter
Initializing Video
Video: 1280x720, windowed
Video Initialized
Renderer: OpenGL version 4.6 (GL_VERSION: 4.6.0 NVIDIA 510.73.05)
Creating FBO...
Starting audio library...
Audio Initialized!
Scanning for Multiverse & Hyperspace patching in ftl.dat
ftl.dat scan detection: Hyperspace.ftl: NO, Multiverse.zip: YES
Resource Preload: 17.030
//////////////////// Skipped all the repeated event crap that doesn't matter
Initializing animations...
Animations Initialized!
Loading Ship Blueprints....
Blueprints Loaded!
//////////////////// Skipped all the repeated event name crap that doesn't matter
Initializing Sound Data....
Generating world...
Loading achievements...
ELF Relocation addr: 0x0000564037188000
ELF Loaded at 0x000056403718d000, 0x0000000000005000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
SEG SIZE: 0x0000000000001000
Fatal Error Failed to find address for function AchievementTracker::LoadAchievementDescriptions:Error parsing hyperspace.xml
Wrong version of Hyperspace detected. Please check that Hyperspace is installed correctly and you are using the correct version of Hyperspace for all of your mods. You can continue, but mods may function incorrectly.Warning: <FTL> not found in data/achievements.xml, adding wrapper tag
Loading score file...
Something went wrong in saving the profile!
Running Game!
Something went wrong in saving the profile!
```

nedsociety:
Well that makes Linux setup unavoidable for me. Wish me luck that WSL2 for Windows 10 can launch FTL with its CPU renderer 😧 

If all of these messes are happening because of the toolchain versions then I'd guess we might need to split the docker images for each platform.

> It's also not clear if you built these for 1.6.12 or 1.6.13 as there is a struct difference between the two and there needs to be the two separate linux binaries for them.

There's no `-DSTEAM_1_6_13_BUILD` set for those builds so I believe they're for 1.6.12.

nedsociety:
Updates:

* Fixed SDL and Discord issues
* Added `-DSTEAM_1_6_13_BUILD` in CMakeLists.txt for now.

Nasa62:
Can we get a shell script to run all the build commands before we merge this in?

It could be separate ones for the different build types or one combined one, but I know it would be nice to occasionally only run one build like Linux x64-1.6.12 or Windows but for releases it would be great to just "build-all"

Or I guess could be a CMake script or something so we can just run `make release-all` or something like that but that would run all the docker commands & everything for us so it's a one button build.

nedsociety:
> Can we get a shell script to run all the build commands before we merge this in?
> 
> It could be separate ones for the different build types or one combined one, but I know it would be nice to occasionally only run one build like Linux x64-1.6.12 or Windows but for releases it would be great to just "build-all"
> 
> Or I guess could be a CMake script or something so we can just run `cmake release-all` or something like that but that would run all the docker commands & everything for us so it's a one button build.

Updated TODO including your suggestion. I'm not exactly sure of the detail of it because some will definitely overlap with Github workflow things so we need to figure it out later.

Nasa62:
> Updated TODO including your suggestion. I'm not exactly sure on the details of it because some will definitely overlap with Github workflow CD. We might need to figure it out later.

Workflow is fine but in the end we need to be able to run it easily on our local machines to generate the builds, and I'd probably want to double check that all builds succeed before a release regardless of if gh-workflow ends up doing the final release build.

nedsociety:
Now all the known bugs are fixed I think this PR is set up to the point it's mergeable.

Some of the recent features require a cached container image. Once the PR gets merged it can be triggered from the github site. Until then the build test should resort to `docker build` method (or [devcontainer CLI](https://code.visualstudio.com/docs/remote/devcontainer-cli)). For the same reason the host-side buildall scripts (`buildall-from-docker.bat/.sh`) won't work until the image caching is done.

PR #35, opened 2022-08-16
Some German image translations

The-Dumb-Dino:
adds the "unlocks disabled" button in german


Nasa62:
Leave out the change to `Hyperspace.ftl` in the future (we'll update it before releases fresh from the `mod_files` folder).

That or the conflict has to be resolved for now, Github should provide instructions for that that I *think* you can do.

We really need to change that file to be auto-built or something as right now when a conflict is caused it's a pain to resolve since that's a binary file just like what's happening here.

PR #43, opened 2022-10-12
zhl changes for clang-12 and higher

nedsociety:
Per @Nasa62's request, taken from #24.

Didn't test if the build is actually working because I don't have a local cbp build setup.


Nasa62:
Closing, already resolved by #24 that happened first.

PR #46, opened 2022-10-19
Pinpoint targeting, cloneSpeedMultiplier, dismissSounds

Vertaalfout:
1. Prevented charge beams from crashing the game
2. Pinpoints (Beams where <length> is 1) now snap to a tile and require only one click
3. Added a crew stat that acts as a multiplier to cloning speed
4. Added the ability to customize the crew dismissal sound in hyperspace.xml


Misc.cpp
```diff
@@ -10,7 +11,10 @@ HOOK_METHOD(CrewEquipBox, RemoveItem, () -> int)
     LOG_HOOK("HOOK_METHOD -> CrewEquipBox::RemoveItem -> Begin (Misc.cpp)\n")
     int ret = super();
     // TODO: R4V3ON Really wants this to be an option to disable in a custom options menu (or maybe in hyperspace.xml)
-    G_->GetSoundControl()->PlaySoundMix("airLoss", -1.f, false);
+    if (CustomOptionsManager::GetInstance()->dismissSound.currentValue==true)
+    {
+        G_->GetSoundControl()->PlaySoundMix(CustomOptionsManager::GetInstance()->dismissSound.currentString, -1.f, false);
```
Nasa62, line 16: We probably want a default value for this so it remains backwards compatible with current Multiverse (since they use their own hyperspace.xml right now) so this currently would cause a breaking change as it would not be defined.

Misc.cpp
```diff
@@ -10,7 +11,10 @@ HOOK_METHOD(CrewEquipBox, RemoveItem, () -> int)
     LOG_HOOK("HOOK_METHOD -> CrewEquipBox::RemoveItem -> Begin (Misc.cpp)\n")
     int ret = super();
     // TODO: R4V3ON Really wants this to be an option to disable in a custom options menu (or maybe in hyperspace.xml)
```
Nasa62, line 13: Remove this TODO since this change accomplishes this.

AdditionalWeaponSlots.cpp
```diff
@@ -28,11 +28,34 @@ HOOK_METHOD_PRIORITY(WeaponControl, RenderAiming, 9999, () -> void)
...
-                if (weapon->targets.size() >= 2)
+                if (weapon->targets.size() >= 2 && weapon->blueprint->length > 1)
                 {
                     RenderBeamAiming(weapon->targets[1], weapon->targets[0], weapon->autoFiring);
                 }
-            }
+                
+                else if(weapon->targets.size() >= 1 && weapon->blueprint->length == 1)
+                {
+                    GL_Primitive *prim;
```
Nasa62, line 38: Please fix the tabbing (indentation) here, this block is aligned incorrectly.

AdditionalWeaponSlots.cpp
```diff
@@ -73,10 +96,32 @@ HOOK_METHOD_PRIORITY(WeaponControl, RenderSelfAiming, 9999, () -> void)
...
-                if (weapon->targets.size() >= 2)
+                if (weapon->targets.size() >= 2 && weapon->blueprint->length > 1)
                 {
                     RenderBeamAiming(weapon->targets[1], weapon->targets[0], weapon->autoFiring);
                 }
+                
+                else if(weapon->targets.size() >= 1 && weapon->blueprint->length == 1)
+                {
+                    GL_Primitive *prim;
+                        if (i<4)
```
Nasa62, line 107: Same here, indentation alignment doesn't match

Nasa62 (changes requested):

Misc.cpp
Vertaalfout: Set default options in CustomOptions.h for backwards compatibility

AdditionalWeaponSlots.cpp
Vertaalfout: Fixed.

Nasa62 (dismissed):
I'm good with this now, but I'd like to see @mathchamp93's approval on it before we merge it just because I'm not super familiar with a lot of the code that was changed here.

TemporalSystem.cpp
```diff
@@ -1211,6 +1211,28 @@ HOOK_METHOD(CloneSystem, OnLoop, () -> void)
 {
     LOG_HOOK("HOOK_METHOD -> CloneSystem::OnLoop -> Begin (TemporalSystem.cpp)\n")
     g_dilationAmount = GetRoomDilationAmount(g_sysDilationRooms, roomId);
+
+    float increment = (G_->GetCFPS()->GetSpeedFactor() * 0.0625);
+    float cloneSpeedMultiplier = 1;
+    auto crewList = G_->GetShipManager(_shipObj.iShipId)->vCrewList;
+    G_->GetCrewFactory()->GetCloneReadyList(crewList,(_shipObj.iShipId==0));
+    int deadCrewCount=G_->GetCrewFactory()->CountCloneReadyCrew(_shipObj.iShipId==0);
```
mathchamp93, line 1219: This is redundant when we're already calling GetCloneReadyList; we can just call size() on the clone list.

RedesignedTooltips.cpp
```diff
@@ -230,8 +230,14 @@ HOOK_METHOD(WeaponBox, GenerateTooltip, () -> std::string)
...
                 currentText = tLib->GetText("beam_length");
-                descText += boost::algorithm::replace_all_copy(currentText, "\\1", std::to_string(bp->length)) + "\n";
-
+                if (bp->length > 1)
+                {
+                    descText += boost::algorithm::replace_all_copy(currentText, "\\1", std::to_string(bp->length)) + "\n";
+                }
+                else
+                {
+                    descText += boost::algorithm::replace_all_copy(currentText, "\\1", "Pinpoint") + "\n";
```
mathchamp93, line 239: Please make this a localizable text (tLib->GetText("foo")).

CustomOptions.h
```diff
@@ -7,6 +7,11 @@ class CustomOptionsManager
...
         rightClickDoorOpening.defaultValue = true;
         rightClickDoorOpening.currentValue = true;
+        
+        dismissSound.defaultValue = true;
+        dismissSound.currentValue = true;
+        dismissSound.defaultString = "airLoss";
+        dismissSound.currentString ="airLoss";
     }
 
     static CustomOptionsManager *GetInstance()
```
mathchamp93, line 17: Do we need to bloat this struct? Perhaps we should change this into a templated struct, i.e., Setting<T>

CustomWeapons.cpp
```diff
@@ -235,6 +235,18 @@ HOOK_METHOD(ProjectileFactory, NumTargetsRequired, () -> int)
     LOG_HOOK("HOOK_METHOD -> ProjectileFactory::NumTargetsRequired -> Begin (CustomWeapons.cpp)\n")
     int ret = super();
 
+    if(blueprint->type == 2)
+    {
+        if(blueprint->length>1)
+        {
+            return 2; //fix for charge beams crashing the game
```
mathchamp93, line 242: Do charge beams work correctly when you fire them for different numbers of charges?

CustomWeapons.cpp
```diff
@@ -247,6 +259,24 @@ HOOK_METHOD(ProjectileFactory, NumTargetsRequired, () -> int)
     return ret;
 }
 
+HOOK_METHOD(ProjectileFactory, Fire, (std::vector<Pointf> &points, int target) -> void)
+{
+    LOG_HOOK("HOOK_METHOD -> ProjectileFactory::Fire -> Begin (CustomWeapons.cpp)\n")
+    if (blueprint->type==2 && blueprint->length==1)
```
mathchamp93, line 265: I'd prefer setting the grid position when aiming rather than when firing but this is acceptable, plus it also covers AI targeting. However the AI always targets the centre of the room so when the AI fires a pinpoint it's always going to snap to the same tile. I would consider adding random tile selection for the AI.

AdditionalWeaponSlots.cpp
```diff
@@ -73,10 +96,32 @@ HOOK_METHOD_PRIORITY(WeaponControl, RenderSelfAiming, 9999, () -> void)
...
             {
-                if (weapon->targets.size() >= 2)
+                if (weapon->targets.size() >= 2 && weapon->blueprint->length > 1)
                 {
                     RenderBeamAiming(weapon->targets[1], weapon->targets[0], weapon->autoFiring);
                 }
+                
+                else if(weapon->targets.size() >= 1 && weapon->blueprint->length == 1)
+                {
+                    GL_Primitive *prim;
```
mathchamp93, line 106: Code duplication; maybe we should turn this into a function/method like with RenderBeamAiming.

mathchamp93 (changes requested):
See comments in files.

RedesignedTooltips.cpp
Vertaalfout: Done

CustomOptions.h
Vertaalfout: Done.

CustomWeapons.cpp
Vertaalfout: The damage doesn't change normally and the animation plays out like a normal beam animation. However, you can modify the damage in lua so that any of the Damage attributes scale with the number of charges.

CustomWeapons.cpp
Vertaalfout: Added random tile selection.

AdditionalWeaponSlots.cpp
Vertaalfout: Resolved code duplication

mathchamp93 (approved):

mathchamp93:
I've made additional changes and I think everything should be good to go now (haven't fully tested everything besides that it compiles/runs with codeblocks and that the pinpoint targeting works for the player).

If you've tested the other features then I'm pretty confident that this can be merged.

FYI when you add methods to the vanilla game's classes you need to add the declarations to the .zhl files. You shouldn't modify the FTLGame.h headers directly as these are generated files.

PR #47, opened 2022-10-21
CI & Release

nedsociety:
Basic build checker.

It can be easily extended to automatically create a release when a tag starting with `v` is pushed (e.g., `v1.2.4`). Would you like it to be implemented?


Nasa62:
When we do a release we need to rebuild `Hyperspace.ftl` zip file, we need to make the final Hyperspace.x.x.x.zip file also that has the binaries in it, then I usually make the tag as part of creating the release on github, although maybe we can do something like a merge to master automatically creates & publishes a release, although I don't know how to from Github's API (although it's probably possible) upload the zip file to the release & publish it all at once.

nedsociety:
> When we do a release we need to rebuild Hyperspace.ftl zip file, we need to make the final Hyperspace.x.x.x.zip file also that has the binaries in it

> although I don't know how to from Github's API (although it's probably possible) upload the zip file to the release & publish it all at once.

These are straightforward.

> then I usually make the tag as part of creating the release on github, although maybe we can do something like a merge to master automatically creates & publishes a release & creates a tag based on the current version number (and maybe errors if we didn't update the version before a recent merge to master, so like we'd update the version in develop at some point or as part of a PR that was breaking etc... to enforce we always remember to do so but not automating it so we don't accidentally push like a breaking change and the automation updates it as a patch),

A popular convention is that humans do merging and tagging, and then CI builds a release whenever a tag is pushed. That's a reasonably safe measure to maximize control over the branch commits (IMO auto-tagging + auto-release for every merge on master sounds awfully error-prone). And also it means you're controlling the version numbers since you're the one naming new tags.

Nasa62:
Maybe it could even update the XML & C files versions for us from the tag name? That would be awesome

Nasa62:
Then it would be just tag in master to generate a release build zip

nedsociety:
That's possible but gonna be a little bit weird, since that means it would push a new commit (that updates the version strings in the sources) and then "move" the tag to the new commit. It'll break sync of work tree between Github and clones.

I'd guess a better way to automate it is either:
1. Make a script (`bump-version.sh`) that auto-update the version strings and then push the tag (example: [script](https://github.com/ftl-mv-translation/mvinstaller/blob/main/snippets/bump_version.py), [workflow](https://github.com/ftl-mv-translation/mvinstaller/blob/main/.github/workflows/release.yml)). You'll merge the master then run this script (e.g., `./bump-version.sh 1.2.4`) to bump & release a version.
2. Just make a release workflow that receives version as input. You'll merge the master then run the workflow from Github site (looks like [this](https://stackoverflow.com/a/62736541/3567518)) to bump & release a version.

nedsociety:
Any direction you'd prefer?

Nasa62:
2 might be nice although honestly the script is probably enough....

I don't exactly switch to master after doing the merge normally though it might be nice to do it all through github, although now that we've got the docker build others can do releases too since they can also build the linux binaries and I can build the windows ones doesn't have to be so coordinated like it used to.

nedsociety:
Done.

> I don't exactly switch to master after doing the merge normally though it might be nice to do it all through github

image

Added a checkbox to merge develop -> master automatically.

---

Note that the first release would require a manual merging. Github is known to hide workflows from UI if a workflow file is not present in the default branch. Once the `release.yml` file is populated by merging, you'll be able to access it from the UI.

Nasa62:
Ned will this break anything if we pull it in but don't start using it for the releases yet?

nedsociety:
Not really. If anything goes south then everyone will get spammed for buildcheck failure whenever they push to the master or develop. That's all.

Nasa62:
> Fixed conflicts and devcontainer build errors.
> 
> Before merging this I'd recommend to [switch the default branch from `master` to `develop`](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/changing-the-default-branch#changing-the-default-branch). It'll make sure that the scheduled and manually triggered workflows are run on develop branch instead of master. And most of the visitors of this repo are more interested in the latest develop codebase than the current stable anyway so it won't hurt anyone.

This looks good, is it okay if I merge it and leave things as master though?
I'm hoping I can get everyone to agree on github-flow soon enough and we just start using master directly and not release partially completed PRs.

.devcontainer/Dockerfile-actual
```diff
@@ -9,13 +9,19 @@ COPY llvm.patch /llvm-build/
 RUN sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list \
     && apt update \
     && apt install -y --no-install-recommends \
-        ca-certificates gnupg build-essential dpkg-dev devscripts locales cmake make python3 zlib1g wget subversion \
-        unzip git \
+        ca-certificates gnupg build-essential dpkg-dev devscripts locales \
+        make python3 zlib1g wget subversion unzip git curl \
```
Nasa62, line 13: subversion is really needed? for what?

Nasa62 (approved):

.devcontainer/Dockerfile-actual
nedsociety: Probably copied from an old build instruction from somewhere. I don't feel it's worth testing as it takes 40+ minutes just to test it 🤣 Anyway it's only there in the first stage when we're compiling the toolchains, and the second stage (`# 2nd image: devcontainer`) resets everything and there's no svn in the final image.

nedsociety:
> This looks good, is it okay if I merge it and leave things as master though?
I'm hoping I can get everyone to agree on github-flow soon enough and we just start using master directly and not release partially completed PRs.

Some notes then:
- The scheduled (monthly) devcontainer build would refer to master. Merging this PR should rebuild the image with `push` trigger, but then at 1st July it will try to rebuild the old version from master. Well the current master version fails to rebuild so it doesn't matter much, but thing will work awkwardly.
- The release workflow won't be shown in the Github actions until the very first time it's merged into master. The next release should be done in the following order:
  1. Merge develop to master by hand, without bumping version. This is a redundant step afterward, but needed to populate the workflow in Github UI.
  2. Run the release workflow.

Nasa62:
@ChronoVortex we should be aware of these steps ned has listed for our next release.
Lets merge this in during our next release.

PR #48, opened 2022-10-28
Custom localization

nedsociety:
## Screenshots

Basic Hyperspace mod:
image

MV:
image

## Summary

Adds a capability to insert a completely new language onto the game. Includes a font override feature that allows you to ship a ttf font file that replaces the entire font set for that language (much like how the current zh-Hans font is implemented in vanilla).

See [this branch](https://github.com/nedsociety/FTL-Hyperspace/tree/feature/custom-localization-with-data/Mod%20Files) for the basic test data.

## Current status

Implementation & test complete on Windows.

Linux signatures and testing are not done - looking for help on these.


libzhlgen/test/functions/win32/1.6.9/freetype.zhl
```diff
@@ -15,6 +15,17 @@ static cleanup uint64_t freetype::easy_printCenter_DO_NOT_USE_DIRECTLY(int fontS
 ".8d4c240483e4f0ff71fc5589e55756535181ec????????8b5110":
 static cleanup __cdecl void freetype::easy_printAutoShrink(int fontId, float x, float y, int maxWidth, bool centered, const std::string& text);
 
+"55660fefc9660fefc089e55389cb83e4f083ec20":
+__thiscall void freetype::font_data::init(freetype::font_data *this, const void *buffer, int bufferSize, unsigned int h, bool glow);
```
Nasa62, line 19: This worked? I thought it didn't work with multiple nested things

nedsociety: Uh, I was dumb when I said it didn't work on discord -- I omitted colon at the end of the sig.

Yeah it worked when I fixed it.

mathchamp93:
One suggestion I have, could we hard-code Hyperspace to automatically add Japanese when running on Windows (1.6.9)? Since that's one thing 1.6.9 is missing that 1.6.12 and later versions had.

nedsociety:
> One suggestion I have, could we hard-code Hyperspace to automatically add Japanese when running on Windows (1.6.9)? Since that's one thing 1.6.9 is missing that 1.6.12 and later versions had.

I considered it as well, there are some thoughts behind not defaulting it:

* As far as I understand how GOG downgrade works there might not be Japanese data at all. Theoretically we may detect if `data/text-ja.xml` exists, but I'm unsure of how to implement it.

* With the current implementation of fontOverride feature we won't be able to properly revive the Japanese font in vanilla. Currently fontOverride is modeled after how Simplified Chinese works: one TTF used for all text. AFAIK Japanese works a bit differently, it actually uses a set of their own bitmap fonts just like how the rest of the languages do. Since I'm ghidra'ing Windows 1.6.9 binary I have no idea how those fonts are initialized and used. Maybe Linux version might shed light on that? Anyway my conclusion was to leave it optional if I can't replicate the font.

Nasa62:
> * As far as I understand how GOG downgrade works there might not be Japanese data at all. Theoretically we may detect if `data/text-ja.xml` exists, but I'm unsure of how to implement it.

Yeah it would be missing from the `ftl.dat` if they added things, I can check if you want to see if it's there in the GoG version but I'm guessing it's not.

> * Since I'm ghidra'ing Windows 1.6.9 binary I have no idea how those fonts are initialized and used. Maybe Linux version might shed light on that? Anyway my conclusion was to leave it optional if I can't replicate the vanilla font.

You can get the Linux binary from Steam or GoG or Humble Bundle if you have those sources, 

```
Also on Steam you can very easily get the Linux version:

You'll need to open the steam console https://www.addictivetips.com/windows-tips/open-steam-console/ then we're following this guide except we're not downloading an older version of a game https://www.reddit.com/r/Steam/comments/611h5e/guide_how_to_download_older_versions_of_a_game_on/

We're interested in the latest Linux depot https://steamdb.info/app/212680/depots/ listed as 221002

So the command should be `download_depot 212680 221002`
Steam should then print out where it put the files
```

Nasa62:
Also I'll get you the Linux signatures on this PR I just haven't had any time yet.

libzhlgen/test/functions/win32/1.6.9/ResourceControl.zhl
```diff
@@ -14,8 +16,8 @@ __thiscall int ResourceControl::RenderImage(ResourceControl *this, GL_Texture* t
 __thiscall int ResourceControl::RenderImageString(ResourceControl *this, std::string& tex, int x, int y, int rotation, GL_Color color, float opacity, bool mirror);
 ".578d7c240883e4f0ff77fc5589e557565383ec5c8b570c8b07f30f1077208b5f048b7708f30f1175d08955d48b5724890424":
 __thiscall GL_Primitive* ResourceControl::CreateImagePrimitiveString(ResourceControl *this, const std::string& tex, int x, int y, int rotation, GL_Color color, float alpha, bool mirror);
-"578d7c240883e4f0ff77fc5589e557565389cb83ec4c8b37":
-__thiscall freetype::font_data &ResourceControl::GetFontData(ResourceControl *this, int fontType, bool unk);
+"57????????83e4f0ff77fc5589e557565389cb83ec4c8b37897dc08b4f048d46fa83f8020f87????????c70709000000b809000000ba01000000be09000000":
+__thiscall freetype::font_data &ResourceControl::GetFontData(ResourceControl *this, int size, bool ignoreLanguage);
```
Nasa62, line 20: Make this same change in the ELF_x86 & ELF_amd64 signatures so that the argument names match

libzhlgen/test/functions/win32/1.6.9/ResourceControl.zhl
```diff
@@ -6,6 +6,8 @@ __thiscall GL_Primitive* ResourceControl::CreateImagePrimitive(ResourceControl *
 __thiscall void ResourceControl::OnInit(ResourceControl *this, int imageSwappingMode);
 "5589e557565389cf31db83e4f083ec408d442430c744240c10000000c744240800000000c744243064617461c74424342e646174894424288b4508":
 __thiscall char* ResourceControl::LoadFile(ResourceControl *this, const std::string& fileName);
+"5589e557565389ce31db83e4f083ec208b45088b7d0cc744240c10000000c7442408000000008b00894424048b81c8000000890424":
```
Nasa62, line 9: Linux 64
`.4156b91000000041554531ed415455` (after `ResourceControl::LoadFile`)

Linux 32
`.5531ed57565383ec2c8b442444` (after `ResourceControl::LoadFile`)

Nasa62: Just putting it here since I had this file up for something else in Ghidra, I think I'll still checkout your branch & add the signatures myself later

libzhlgen/test/functions/win32/1.6.9/ResourceControl.zhl
```diff
@@ -14,8 +16,8 @@ __thiscall int ResourceControl::RenderImage(ResourceControl *this, GL_Texture* t
 __thiscall int ResourceControl::RenderImageString(ResourceControl *this, std::string& tex, int x, int y, int rotation, GL_Color color, float opacity, bool mirror);
 ".578d7c240883e4f0ff77fc5589e557565383ec5c8b570c8b07f30f1077208b5f048b7708f30f1175d08955d48b5724890424":
 __thiscall GL_Primitive* ResourceControl::CreateImagePrimitiveString(ResourceControl *this, const std::string& tex, int x, int y, int rotation, GL_Color color, float alpha, bool mirror);
-"578d7c240883e4f0ff77fc5589e557565389cb83ec4c8b37":
```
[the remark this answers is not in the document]
nedsociety: image

(sorry for the color issue -- SecureUxTheme sometimes messes up some apps)

libzhlgen/test/functions/win32/1.6.9/ResourceControl.zhl
nedsociety: I'll address the previous comment and rebase this PR today (approx 12 hrs later I guess). Feel free to checkout after that if you'd want to avoid merge conflicts.

libzhlgen/test/functions/win32/1.6.9/ResourceControl.zhl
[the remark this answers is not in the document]
Nasa62: If there's no `.` before the signature it'll match the first address restarting from the 0th position of the search in the application. If there was a `.` it starts from the match of the last address.

The last function match address is the address of the last signature scanned forwards to the first `RET` instruction, unless that previous signature has `!` flag in which case the pointer is the matched function address + the signature byte length.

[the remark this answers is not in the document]
Nasa62: Basically without a `.` it'll match the first signature found with that pattern, so the lowest address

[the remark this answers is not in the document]
Nasa62: The order of signature scanning is 100% deterministic, it used to only be deterministic per ZHL file but I tweaked it when I started the Linux port so it was always deterministic and reads all ZHL files in sorted order, so alphabetical in this case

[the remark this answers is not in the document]
nedsociety: Hmm. Alright I'll revert this change then.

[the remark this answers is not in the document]
Nasa62: I haven't checked that that signature *is correct* yet but I assume it probably was.

If we never actually called that method it's possible the signature was wrong.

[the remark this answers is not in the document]
nedsociety: > If we never actually called that method it's possible the signature was wrong.

Well there's no usage of it besides being exposed from swig so I'll just let it as-is for now then.

nedsociety:
Rebased to the latest and changed argnames

libzhlgen/test/functions/win32/1.6.9/ResourceControl.zhl
Nasa62: Were these signatures ever added? I don't see them.

libzhlgen/test/functions/win32/1.6.9/OptionsScreen.zhl
```diff
@@ -4,3 +4,8 @@ __thiscall void OptionsScreen::OnInit(OptionsScreen *this);
 __thiscall void OptionsScreen::OnLoop(OptionsScreen *this);
 "57660f76c08d7c240883e4f0ff77fc5589e557565381ec9c0400008b07898d70fbffff0f1181f8090000c681cc02000000":
 __thiscall void OptionsScreen::Open(OptionsScreen *this, bool mainMenu);
+"57????????83e4f0ff77fc5589e557565389ce8d4d9081ecac000000":
```
Nasa62, line 7: This should be in its own file `LanguageChooser.zhl` to be consistent with how we were structuring all the other files.

`LanguageChooser::OnRender`
Linux:
64: `4157bef001000041564155`
32: `5557565331db83ec6c`

libzhlgen/test/functions/win32/1.6.9/OptionsScreen.zhl
```diff
@@ -4,3 +4,8 @@ __thiscall void OptionsScreen::OnInit(OptionsScreen *this);
...
+__thiscall void LanguageChooser::OnRender(LanguageChooser *this);
+
+"!83ec1ca1(???????\?)c1e30301d8":
```
Nasa62, line 10: Looks like this might still be part of `LanguageChooser` not sure if it's part of `OptionsScreen` as it wasn't organized that way like the other globals were when I searched in Ghidra but I'm not 100% certain.


Anyways
Linux 64: (has to use offsetVariable because of RIP-relative addressing in this match) 
This signature is technically non-unique, but both matches are for languageList anyways :smile:
```
"!4889ee480335(???????\?)":
offsetVariable std::vector<std::String>* Global_OptionsScreen_languageList;
```

Linux 32: (direct match, use same declaration as Win32)
`!8d348500000000a1(???????\?)`


^^^ Both of these matches are in `LanguageChooser::OnLoop` which in Linux 32 & 64 is *before* `LanguageChooser::OnRender`
If you wanted ideal optimization, these could be declared before `LanguageChooser::OnRender` and you could use `.` at the start of the `OnRender` methods so that you only have to scan ahead from the point of this variable match and not from the start of the entire file.

libzhlgen/test/functions/win32/1.6.9/freetype.zhl
```diff
@@ -15,6 +15,17 @@ static cleanup uint64_t freetype::easy_printCenter_DO_NOT_USE_DIRECTLY(int fontS
 ".8d4c240483e4f0ff71fc5589e55756535181ec????????8b5110":
 static cleanup __cdecl void freetype::easy_printAutoShrink(int fontId, float x, float y, int maxWidth, bool centered, const std::string& text);
 
+"55660fefc9660fefc089e55389cb83e4f083ec20":
```
Nasa62, line 18: `init` comes before `clean` and that comes long before `easy_measurePrintLines` in Linux, as such, this signature needs to be at the top of the file and **`easy_measurePrintLines` needs to have a `.` added to its signature for optimization so it doesn't restart the scan.**
The order on Linux needs to be:
1. init
2. clean
3. easy_measurePrintLines & everything else
4. font_height
5. font_baseline

Linux:
`freetype::font_data::init`
64: `89c9534889fbf3480f2ac14889f731c989d631d2`
32: `5383ec288b44243c8b5c24300fb7d0c1e810`

libzhlgen/test/functions/win32/1.6.9/freetype.zhl
```diff
@@ -15,6 +15,17 @@ static cleanup uint64_t freetype::easy_printCenter_DO_NOT_USE_DIRECTLY(int fontS
...
+__thiscall void freetype::font_data::init(freetype::font_data *this, const void *buffer, int bufferSize, unsigned int h, bool glow);
+"5589e583e4f083ec108b4104890424e8????????c9c3":
```
Nasa62, line 20: Linux (must be after `init` is declared)
`freetype::font_data::clean`
64: `!.8b7f04e9`
32: `!.8b4424048b400489442404`

libzhlgen/test/functions/win32/1.6.9/freetype.zhl
```diff
@@ -15,6 +15,17 @@ static cleanup uint64_t freetype::easy_printCenter_DO_NOT_USE_DIRECTLY(int fontS
...
+__thiscall void freetype::font_data::clean(freetype::font_data *this);
+"5589e5565383e4f083ec308b4d08f30f104d0c85c97e??3b0d????????7d??803d????????008b15????????74??8b1d????????39d97c??8d349d000000008b143285d274??89c829d88db60000000039c389c17f??8b143229d885d275??660fefc0eb??8b1c8a660fefc085db74??0f2ec173??0f2ec876??0f2e4b247a??75??f30f10432c":
```
Nasa62, line 22: `font_height` must be declared ***before*** `font_baseline` in Linux
`font_height` should be declared ***after*** `freetype::easy_printAutoShrink`

`font_height` Linux 64: `.85ff530f8e`
`font_height` Linux 32: `.57565383ec208b4c2430f30f1044243485c9`
`font_baseline` Linux 64: `.85ff530f8e`
`font_baseline` Linux 32: `.57565383ec208b4c2430f30f1044243485c9`

libzhlgen/test/functions/win32/1.6.9/freetype.zhl
```diff
@@ -15,6 +15,17 @@ static cleanup uint64_t freetype::easy_printCenter_DO_NOT_USE_DIRECTLY(int fontS
...
+cleanup __cdecl float font_baseline(int font_id, float size);
+"5589e5565383e4f083ec308b4d08f30f104d0c85c97e??3b0d????????7d??803d????????008b15????????74??8b1d????????39d97c??8d349d000000008b143285d274??89c829d88db60000000039c389c17f??8b143229d885d275??660fefc0eb??8b1c8a660fefc085db74??0f2ec173??0f2ec876??0f2e4b247a??75??f30f104328":
+cleanup __cdecl float font_height(int font_id, float size);
+
+"!89108b559cc705(???????\?)000000408950048b55a0":
+float *Global_freetype_sil_freetype_outline;
```
Nasa62, line 28: In Linux this comes after `font_baseline` (so already fine if other comments were followed)
Linux 64:
```
"!.f30f5905(???????\?)4531c031c9f30f2df0":
offsetVariable float* Global_freetype_sil_freetype_outline;
```

Linux 32: (direct address match, use like Win32)
`!.f30f5905(???????\?)c744240801000000f30f2dc0`

Nasa62 (changes requested):
More signature information

The-Dumb-Dino:
im curious, what is holding it back from being merged? are there still missing signatures?

Nasa62:
> im curious, what is holding it back from being merged? are there still missing signatures?

I think the signatures were provided above here, they just need to be added & also ned needs to bring it up to speed with the latest commits & re-generate the FTLGame files.

ChronoVortex (approved):
As long as the vanilla `ResourceControl::PreloadResources` function is only called once at the start of the game, which I'm pretty sure it is, there shouldn't be an issue with turning the `FreeTypeFont` references in `g_customFontOverrides` into pointers. Though, if there do turn out to be memory leak issues when 1.10.0 is out, this would be a place to check.

PR #64, opened 2022-11-16
fixed bugs when customSystem info for engine was impartially/not decl…

Vertaalfout:
…ared

### DIFF
--- a/CustomSystems.cpp
+++ b/CustomSystems.cpp
@@ -151,9 +151,16 @@ HOOK_METHOD(EngineSystem, GetDodgeFactor, () -> int)
     else
     {
         auto levelVector = CustomSystemManager::GetInstance()->engineDodgeLevels.systemLevels;
-         //TODO: Make sure it defaults properly if none of this information is declared.
-        int overrideLevel  = (this->_shipObj.iShipId == 0) ? levelVector[effectivePower-1].first : levelVector[effectivePower-1].second;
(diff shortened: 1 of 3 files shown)


CustomSystems.cpp
```diff
@@ -151,9 +151,16 @@ HOOK_METHOD(EngineSystem, GetDodgeFactor, () -> int)
     else
     {
         auto levelVector = CustomSystemManager::GetInstance()->engineDodgeLevels.systemLevels;
-         //TODO: Make sure it defaults properly if none of this information is declared.
-        int overrideLevel  = (this->_shipObj.iShipId == 0) ? levelVector[effectivePower-1].first : levelVector[effectivePower-1].second;
-        if (overrideLevel == -2147483648 || (effectivePower > levelVector.size())) {return super();} //when a <level> node doesn't have a defined evasionBoost, it will default to -2147483648 for both the player and enemy
+        int overrideLevel;
+        if (effectivePower > levelVector.size())
+        {   
+            overrideLevel = -2147483648;
```
Nasa62, line 157: Where does the magic number come from?

Vertaalfout: I made it so that whenever information was missing from the parsed data or anything, it was replaced with this number, which is always handled to mean "use the default information" 

Should I have used something like INT_MIN instead?

Nasa62: It doesn't need to be `INT_MIN` but a magic number like this shouldn't be repeated manually over and over again, make a global static constant variable or something just so it's something referenced and so that it's clear they're related and it's not just pulled out of thin air.

PR #73, opened 2023-01-13
exposed ShipGraph for LUA

ChronoVortex:


lua/modules/hyperspace.i
```diff
@@ -13,6 +13,8 @@
 #include "Projectile_Extend.h"
 %}
 
+%feature("flatnested");
```
Nasa62, line 16: What does this do?

ChronoVortex: This is something Vert suggested to me:
`If the target language doesn't support nested classes directly, or the support is not implemented in the language module (like for Python currently), then the visible nested classes are moved to the same name space as the containing class (nesting hierarchy is "flattened"). The same behaviour may be turned on for C# and Java by the %feature ("flatnested"); If there is a class with the same name in the outer namespace the inner class (or the global one) may be renamed or ignored:`
Basically it makes it so that the nested structures inside `Globals` can be exposed

Nasa62 (approved):

PR #74, opened 2023-01-17
bug fixes

ChronoVortex:
- ~~fixed infinite loop due to unsigned int~~
- added missing return to hacking drone collision check
- drone derived classes were broken, replaced with SpaceDrone
- fixed SHIP prefix for HasEquipment

### DIFF
--- a/lua/LuaLibScript.cpp
+++ b/lua/LuaLibScript.cpp
@@ -33,14 +33,14 @@ void LuaLibScript::LoadTypeInfo()
     types.pProjectile[6] = SWIG_TypeQuery(this->m_Lua, "PDSFire *");
     types.pShipManager = SWIG_TypeQuery(this->m_Lua, "ShipManager *");
 
-    types.pSpaceDroneTypes[0] = SWIG_TypeQuery(this->m_Lua, "DefenseDrone *");
-    types.pSpaceDroneTypes[1] = SWIG_TypeQuery(this->m_Lua, "CombatDrone *");
+    types.pSpaceDroneTypes[0] = SWIG_TypeQuery(this->m_Lua, "SpaceDrone *");
+    types.pSpaceDroneTypes[1] = SWIG_TypeQuery(this->m_Lua, "SpaceDrone *");
     types.pSpaceDroneTypes[2] = nullptr;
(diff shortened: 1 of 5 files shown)


lua/LuaLibScript.h
```diff
@@ -34,7 +34,7 @@ class LuaLibScript
         bool call_on_internal_chain_event_callbacks(InternalEvents::Identifiers, int nArg=0, int nRet=0);
 
         int call_on_render_event_pre_callbacks(RenderEvents::Identifiers, int nArg);
-        void call_on_render_event_post_callbacks(RenderEvents::Identifiers, unsigned int idx, int nArg);
```
Nasa62, line 37: Why did you remove the `unsigned`? This index value is never negative, it's an array index, it must always be `>= 0`.

lua/LuaLibScript.cpp
```diff
@@ -237,7 +237,7 @@ int LuaLibScript::call_on_render_event_pre_callbacks(RenderEvents::Identifiers i
 
     return idx;
 }
-void LuaLibScript::call_on_render_event_post_callbacks(RenderEvents::Identifiers id, unsigned int idx, int nArg)
```
Nasa62, line 240: See other comment about `unsigned`

Nasa62 (changes requested):
Changes about `unsigned`

lua/LuaLibScript.h
Nasa62: Or I guess, describe how this causes an infinite loop, because I don't understand what the heck you mean.

ChronoVortex: The loop continues as long as idx >= 0, which is always true for an unsigned int.

Nasa62: I'll need to look at my code, but this doesn't seem like the ideal way to fix it if it's a logic bug, the API should remain that it's an index number of an array which is bounded by 0.

ChronoVortex: I was kind of afraid to touch the logic, I didn't completely get it. Shall I undo this change for this PR?

ChronoVortex: I undid it

Nasa62 (approved):

PR #76, opened 2023-01-31
overwrote original OnRenderBase function

ChronoVortex:
Thruster animations now render under the floor image. Also improved the cloaking alpha transition for thruster animations and hull sprite.


Nasa62:
Does this remain backwards compatible with the existing use of engine sprites?

ChronoVortex:
I believe so, definitely for Multiverse. As far as I know no other HS mod or MV addon depends on the thruster anims rendering over the floor image.

Nasa62 (changes requested):
Add documentation to hyperspace.xml about using the new engine sprites rendering layer

*Optional: additionally provide updates for the wiki about this feature

Nasa62:
Okay without protest from addon developers seems fine.

But is there sufficient information about using this in hyperspace.xml or the other example files of the mod-files folder (or optionally the wiki) to show how to use thruster animations in general?
Not necessarily from your change but was there even information there before? If not, I'd like to fill the documentation gap as part of this PR. If it's already well documented let me know and we're good to go.

ChronoVortex:
Added

Nasa62 (approved):

PR #78, opened 2023-02-02
Experimental features and minor fixes

ChronoVortex:
CommandConsole.cpp: Fixed format string crash with strings like "%s" .(This was an issue when trying to use string.format in lua).

CustomAugments.cpp, lua/InternalEvents.h:
Added callback for GetAugmentationValue (This is useful for scaling augment effects and the like).

CustomDamage.cpp: Fixed bug where bombs spawned via SpaceManager::CreateBomb would loop their animation.

Global.cpp:
Added flag to MessageBoxW so that the error message will reliably spawn in front of the game window.

lua/luaGlobal.h:
Changed the lua log function so that invalid arguments cause a lua error instead of an assertation fail.

lua/LuaScriptInit.cpp, lua/LuaScriptInit.h, lua/modules/rapidxml.i,
CMakeLists.txt,
FTL_HyperspaceGCC.cbp:

~~Added RapidXML module to swig for XML parsing. (Briefly tested, functional).~~

lua/RenderEvents.cpp, lua/RenderEvents.h:

Added SHIP_SPARKS render callback with iShipId argument. (Useful for dynamic/automatic room images, such as system location indications).

lua/modules/hyperspace.i:

Exposed ResourceControl::LoadFile (to be used with xml parsing) and SystemTemplate.


lua/modules/rapidxml.i
```diff
@@ -0,0 +1,22 @@
+%module RapidXML
```
Nasa62, line 1: Vertall did this right?

Why even expose rapidxml? What purpose?

ChronoVortex: Exposing RapidXML allows you to read the XML in LUA and do things based on it, so you can do things like add logic for a custom tag.

lua/RenderEvents.cpp
```diff
@@ -51,6 +51,17 @@ HOOK_METHOD(CommandGui, RenderPlayerShip, (Point &shipCenter, float jumpScale) -
     Global::GetInstance()->getLuaContext()->getLibScript()->call_on_render_event_post_callbacks(RenderEvents::LAYER_PLAYER, std::abs(idx), 0);
 }
 
+HOOK_METHOD_PRIORITY(Ship, OnRenderSparks, -1000, () -> void)
```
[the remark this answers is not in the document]
ChronoVortex: Ran the script

CustomAugments.cpp
```diff
@@ -487,6 +487,24 @@ HOOK_METHOD_PRIORITY(ShipObject, GetAugmentationValue, 1000, (const std::string&
     return ret;
 }
 
+HOOK_METHOD_PRIORITY(ShipObject, GetAugmentationValue, 500, (const std::string& name) -> float)
+{
+    LOG_HOOK("HOOK_METHOD_PRIORITY -> ShipObject::GetAugmentationValue -> Begin (CustomAugments.cpp)\n")
+    float ret = super(name);
+
+    auto context = Global::GetInstance()->getLuaContext();
+    SWIG_NewPointerObj(context->GetLua(), G_->GetShipManager(this->iShipId), context->getLibScript()->types.pShipManager, 0);
```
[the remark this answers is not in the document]
ChronoVortex: Fixed

lua/RenderEvents.cpp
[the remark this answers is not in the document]
ChronoVortex: And again in the lua folder

CommandConsole.cpp
```diff
@@ -23,7 +23,7 @@ HOOK_METHOD(MouseControl, OnRender, () -> void)
 
 bool CommandConsole::RunCommand(CommandGui *commandGui, const std::string& cmd)
 {
-    hs_log_file(cmd.c_str());
+    hs_log_file("%s", cmd.c_str());
```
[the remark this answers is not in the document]
ChronoVortex: Not sure how to address this myself, reverted for now

lua/RenderEvents.cpp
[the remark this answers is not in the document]
ChronoVortex: There's already an OnRenderSparks hook in CustomShips, the priority here is to ensure it's executed last.

CustomAugments.cpp
```diff
@@ -487,6 +487,24 @@ HOOK_METHOD_PRIORITY(ShipObject, GetAugmentationValue, 1000, (const std::string&
     return ret;
 }
 
+HOOK_METHOD_PRIORITY(ShipObject, GetAugmentationValue, 500, (const std::string& name) -> float)
```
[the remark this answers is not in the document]
ChronoVortex: The GetAugmentationValue hook that overwrites the vanilla function has a priority of 1000. From what I can tell, since this has a priority of 500, any modifications it makes to the augment values will come after that base function. Vert says the reason is probably more sophisticated than that though, so you'll probably wanna hear from them.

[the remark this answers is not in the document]
ChronoVortex: No vert has given me a more detailed explanation. If you have this code:
```c++
HOOK_METHOD_PRIORITY(Class, Method, 1000, () -> void)
{
  print("Priority 1000 Before. \n");
  super();
  print("Priority 1000 After. \n");
}
HOOK_METHOD_PRIORITY(Class, Method, 500, () -> void)
{
  print("Priority 500 Before. \n");
  super();
  print("Priority 500 After. \n");
}```
you get this output:
```
Priority 500 Before.
Priority 1000 Before.
Priority 1000 After.
Priority 500 After.```
So the old GetAugmentationValue runs its code, this new GetAugmentationValue gets it's return value and then it runs its code.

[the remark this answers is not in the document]
ChronoVortex: So vert has given me a more detailed explanation. If you have this code, for example:
```c++
HOOK_METHOD_PRIORITY(Class, Method, 1000, () -> void)
{
  print("Priority 1000 Before. \n");
  super();
  print("Priority 1000 After. \n");
}
HOOK_METHOD_PRIORITY(Class, Method, 500, () -> void)
{
  print("Priority 500 Before. \n");
  super();
  print("Priority 500 After. \n");
}
```
you would get this output:
```
Priority 500 Before.
Priority 1000 Before.
Priority 1000 After.
Priority 500 After.
```
The way this is set up, the old GetAugmentationValue hook runs its code, then this new GetAugmentationValue hook gets its return value and runs *its* code.

lua/modules/rapidxml.i
Nasa62: Yeah but... I'd rather expose a Lua XML parser to Lua so it's easier to use if we're going to allow it, then maybe we can have a function to convert between RapidXML and whatever library format because that would probably be easier to deal with for developers. We only use RapidXML because FTL uses it internally and we had to hook some things that used it.

RapidXML is *really, really, clunky* plus not sure the security issues of that or not, needs more investigation if that's safe, it might be just fine.

ChronoVortex: I'm pretty sure vert was able to confirm that exposing RapidXML doesn't let you do anything like read files outside of FTL's data, so I don't think security is an issue. I also don't really see the issue with using it over a native LUA parser, it's pretty smooth with just a couple helper functions. Plus I've already got the [Vertex mod](https://github.com/ChronoVortex/FTL-HS-Vertex) built on top of it and it would be a pain in the butt to change it :P

ChronoVortex: Updates from vert:
- RapidXML is preferable to a LUA parser because it would be extremely difficult to set a LUA parser up to have parity with RapidXML. You don't want a situation where RapidXML is okay with something that the LUA parser isn't.
- Not certain about security, may need some additional testing.

ChronoVortex: Alright, I've been thinking about this and I don't believe exposing RapidXML has any security issues. The only potential security concern I can think of, which has nothing to do with RapidXML, is exposing `Resources:LoadFile` allowing you to `../` into directories you shouldn't be able to read. It doesn't though, I've tested it and it only seems to let you work inside the `dat`.

CustomAugments.cpp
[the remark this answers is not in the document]
Nasa62: Yes but if you don't supply a priority then you're just in the default ordering.

We try to only supply a priority if it's 100% important that we be the *first* or *last* hook in the chain of hooks, we try not to manually specify priorities for hooks because that would get really painful to maintain.

lua/RenderEvents.cpp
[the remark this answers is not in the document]
Nasa62: Is this defined *after* the other hook? If so, take out the priority as the order of definition already implies a priority.

Or if needs to be before just that one hook then it can be defined before it.

If it needs to be *globally* before or after all hooks for that method, that's when priority is used.

lua/modules/rapidxml.i
Nasa62: Can we split this part out into a separate PR? I think I want to try having us work on a layer that uses a Lua XML parser, yes maybe it has a method to convert to RapidXML. I think that would be a better approach?

ChronoVortex: If you don't wanna expose RapidXML at all shouldn't I just remove it entirely instead of moving it to a different PR?

lua/RenderEvents.cpp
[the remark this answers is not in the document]
ChronoVortex: Looking at the CMake file, it looks like this source file _is_ loaded after the one with the other hook. Fixed.

CustomAugments.cpp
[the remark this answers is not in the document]
ChronoVortex: Fixed

lua/modules/rapidxml.i
ChronoVortex: Removed RapidXML

Global.cpp
```diff
@@ -90,7 +90,7 @@ void ErrorMessage(const char *msg)
 {
     #ifdef _WIN32
         std::wstring utf16String = ConvertToUtf16(msg, CP_UTF8);
-        MessageBoxW(NULL, utf16String.c_str(), L"Error", MB_ICONERROR);
+        MessageBoxW(NULL, utf16String.c_str(), L"Error", MB_ICONERROR | MB_SETFOREGROUND);
     #elif defined(__linux__)
         SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, "Error", msg, NULL);
```
Nasa62, line 95: I'd like to find if there's a way do do the foreground in SDL but I don't know if there is. As SDL works cross-platform by MessageBoxW is Windows only so might be nice to clean this up at some point and also fix the bug cross-platform.

CustomAugments.cpp
mathchamp93, line 1: What preempt could theoretically do is if the callback was before super then preempt would skip the super. But that doesn't really make sense here since we're manipulating a number, so having the super first makes more sense (then the callback can see the number and adjust it).

Another alternative would be to have some way to make the super callable from the callback or even rewrite it, that way the super call can be avoided if completely redirecting an augment effect away from augments. However in most cases with this callback I'd imagine checking the augments normally first would still happen, then you'd do extra manipulation.

AugmentValues.cpp
mathchamp93, line 1: We should call GetAugmentationValue only once here.

Augments.cpp
mathchamp93, line 1: We should call GetAugmentationValue only once here.

CustomAugments.cpp
```diff
@@ -487,6 +487,24 @@ HOOK_METHOD_PRIORITY(ShipObject, GetAugmentationValue, 1000, (const std::string&
     return ret;
 }
 
+HOOK_METHOD(ShipObject, GetAugmentationValue, (const std::string& name) -> float)
+{
+    LOG_HOOK("HOOK_METHOD -> ShipObject::GetAugmentationValue -> Begin (CustomAugments.cpp)\n")
+    float ret = super(name);
+
+    auto context = G_->getLuaContext();
+    SWIG_NewPointerObj(context->GetLua(), G_->GetShipManager(this->iShipId), context->getLibScript()->types.pShipManager, 0);
```
mathchamp93, line 496: Do we want to have the ShipManager be the argument, or would just the ship ID or the ShipInfo be better?

lua/modules/hyperspace.i
```diff
@@ -2173,6 +2173,8 @@ playerVariableType playerVariables;
 %rename("%s") ResourceControl::RenderImage;
 %rename("%s") ResourceControl::RenderImageString;
 
+%rename("%s") ResourceControl::LoadFile;
```
mathchamp93, line 2176: Do we want this exposed? This opens a resource file and returns its contents (I think) as a char* (or char array).

Also, LoadFile allocates new memory (for the returned string); how are we managing this? I don't want to expose a method if it is going to leak memory every time it is called.

mathchamp93 (changes requested):
General comment: It doesn't affect your changes but some vanilla augments (ones that have a boolean function like DNA bank, hacking stun, etc.) have a value of 0, and the game just checks if those augments are present. Eventually should add callbacks for methods such as HasEquipment and HasAugmentation.

CustomAugments.cpp
ChronoVortex: ShipManager is better since its consistent with other callbacks

AugmentValues.cpp
ChronoVortex: Fixed

Augments.cpp
ChronoVortex: Fixed

lua/modules/hyperspace.i
ChronoVortex: Ran a quick test to check if `%newobject ResourceControl::LoadFile;` plugs the leak and it does. Loading blueprints every frame without it causes the game to crash in a few seconds. Loading blueprints every frame _with_ it does not crash the game, and checking the memory usage in task manager confirms that it doesn't increase.

mathchamp93: Okay. Let's hope the delete[] on the LoadFile result works properly on both Windows and Linux.

lua/RenderEvents.cpp
```diff
@@ -51,6 +51,17 @@ HOOK_METHOD(CommandGui, RenderPlayerShip, (Point &shipCenter, float jumpScale) -
     Global::GetInstance()->getLuaContext()->getLibScript()->call_on_render_event_post_callbacks(RenderEvents::LAYER_PLAYER, std::abs(idx), 0);
 }
 
+HOOK_METHOD(Ship, OnRenderSparks, () -> void)
+{
+    LOG_HOOK("HOOK_METHOD_PRIORITY -> Ship::OnRenderSparks -> Begin (RenderEvents.cpp)\n")
+    auto context = Global::GetInstance()->getLuaContext();
+    lua_pushinteger(context->GetLua(), this->iShipId);
```
mathchamp93, line 58: Why not push "this" itself here? Since we already have the pointer.

If "Ship" isn't exposed we can expose it (and at least its iShipId property).

mathchamp93 (changes requested):
I'm going to let you expose rapidxml since I actually like the intended usage (being able to define custom xml files for other mods to add to in order to interface with yours without necessarily needing to learn lua) now that I look at it. I plan to create a class that inherits the rapidxml document object but wraps it with the parsed string itself, then add some helper methods (e.g., to combine LoadFile and parsing into a single method without having to pass a string back and forth between lua and c++).

ChronoVortex:
> I'm going to let you expose rapidxml since I actually like the intended usage (being able to define custom xml files for other mods to add to in order to interface with yours without necessarily needing to learn lua) now that I look at it. I plan to create a class that inherits the rapidxml document object but wraps it with the parsed string itself, then add some helper methods (e.g., to combine LoadFile and parsing into a single method without having to pass a string back and forth between lua and c++).

Yay! I've still got rapidxml in another branch, I'll open it as a separate PR

lua/RenderEvents.cpp
ChronoVortex: Done, `Ship` is exposed already

CommandConsole.cpp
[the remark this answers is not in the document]
mathchamp93: I would honestly worry more about hslua_g_printMessage than RunCommand in that case. RunCommand is going to be something the player typed in whereas hslua_g_printMessage can be anything hidden in a lua script.

mathchamp93 (approved):

CommandConsole.cpp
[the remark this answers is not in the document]
mathchamp93: Actually, hs_log_file("%s", cmd.c_str()) would be safe and hs_log_file(cmd.c_str()) would be vulnerable. Though the player would be hacking their own game here since they're the one putting in commands.

PR #80, opened 2023-02-13
fixed hacking sound loop when system not installed

ChronoVortex:
If you try to apply a hack effect through LUA, the hacking sound loop while in combat is incorrect. These changes fix that issue. ~~Currently, only Windows `HackingSystem::SoundLoop` and `SoundControl::UpdateSoundLoop` instruction signatures exist, still needs Linux instruction signatures.~~


CustomShips.cpp
```diff
@@ -472,6 +472,27 @@ HOOK_METHOD(ShipManager, OnLoop, () -> void)
...
     }
+    
+    bool hackSoundLoop = false;
+    if (hackingSystem)
+    {
+        hackSoundLoop = hackingSystem->SoundLoop();
+    }
+    if (!hackSoundLoop)
+    {
+        ShipManager* otherShip = G_->GetShipManager((iShipId + 1)%2);
```
Nasa62, line 483: iShipID + 1 modulo 2?

Why not just `G_->GetShipManager(1)` since that's the enemy ship.
`0` is the player ship, probably ideal if we had an enumeration for iShipId but it's whatever since it commonly comes from the game.

------

Or I see you're just trying to get the other one, then in that case *yes the modulo works* but this makes for less readable code since it's not needed, a simple ternary might be nicer for readability like:

`iShipId == PLAYER ? ENEMY : PLAYER` (assuming we had an enumeration for it, which I don't think we do)
so
`iShipId == 1 ? 0 : 1` otherwise but at least it's clear we're only selecting one or the other ship?


I dunno, it just *feels* wrong to me like the modulo is because of the lack of an enumeration and if there was an enumeration the module would be nuts to do, so I feel it would be cleaner to treat it as if it was and maybe we can clean that up later to use an enumeration (or do that now) so it's clearer to everyone reading it that we're just selecting one of two ships.

CustomShips.cpp
```diff
@@ -472,6 +472,27 @@ HOOK_METHOD(ShipManager, OnLoop, () -> void)
...
+        if (otherShip) for (auto system : otherShip->vSystemList) if (system->bUnderAttack && system->iHackEffect == 2)
+        {
+            hackSoundLoop = true;
+            break;
+        }
+        if (!hackSoundLoop) for (auto system : vSystemList) if (system->bUnderAttack && system->iHackEffect == 2)
```
Nasa62, line 489: it's legal, but man I hate this, combining single-line if's and for's with a multi line brace and then all shoved on a single line with no tabbing over on multiple lines...

maybe at least space it on separate lines? I know we don't have any linting rules but... I dunno, this bugs me and probably only me.

```c++
if(!hackSoundLoop)
    for(blah)
        if(bleh) {
        }
```

I'll still approve even without any changes to this, it's just a bugs me thing.

libzhlgen/test/functions/win32/1.6.9/HackingSystem.zhl
```diff
@@ -2,3 +2,5 @@
 __thiscall void HackingSystem::BlowHackingDrone(HackingSystem *this);
 "578d7c240883e4f0ff77fc5589e557565389cb83ec4ce8????????80bbe4010000000f84????????80bbc0010000000f84????????8b8bac070000":
 __thiscall void HackingSystem::OnLoop(HackingSystem *this);
+"80b9c00100000075??31c0c38d7426005589e55389cb83e4??e8????????83f8??74??":
+__thiscall bool HackingSystem::SoundLoop(HackingSystem *this);
```
Nasa62, line 6: Need signatures for Linux before we can merge, are you able to get the Linux binary from Steam?

If you have GoG or Humble Bundle copies it's even easier you can get it directly from their websites.

### Here's the instructions for getting it from Steam:
Also on Steam you can very easily get the Linux version:

You'll need to open the steam console https://www.addictivetips.com/windows-tips/open-steam-console/ then we're following this guide except we're not downloading an older version of a game https://www.reddit.com/r/Steam/comments/611h5e/guide_how_to_download_older_versions_of_a_game_on/

We're interested in the latest Linux depot https://steamdb.info/app/212680/depots/ listed as 221002

So the command should be `download_depot 212680 221002`
Steam should then print out where it put the files

Nasa62 (changes requested):

libzhlgen/test/functions/win32/1.6.9/HackingSystem.zhl
ChronoVortex: I was able to get the Linux binary from Humble Bundle I believe, still need to get the one from Steam. Will work on cracking that open and grabbing the signatures.

ChronoVortex: Added Linux sigs

CustomShips.cpp
ChronoVortex: Fixed

CustomShips.cpp
ChronoVortex: Fixed (pain)

libzhlgen/test/functions/ELF_amd64/1.6.13/HackingSystem.zhl
```diff
@@ -2,3 +2,5 @@
 cleanup __amd64 void HackingSystem::BlowHackingDrone(HackingSystem *this);
 "55534889fb4883ec38e8????????80bb7002000000":
 cleanup __amd64 void HackingSystem::OnLoop(HackingSystem *this);
+"5380bf44020000????89fb75??31c05bc30f1f8000000000e8????????83f8??75??0fb683b0050000":
```
Nasa62, line 5: Wildcard's more than needed, this works but so does:
`"5380bf44020000004889fb750b31c05bc30f1f8000000000e8????????83f8ff"`

Although it is cheating a little because I know from comparing both that I can skip the wildcarding of the `JNZ` so:
`"5380bf44020000004889fb75??31c05bc30f1f8000000000e8????????83f8ff"`
Is technically the most correct of just skipping the addresses with the `A` button in Ghidra's instruction pattern search

This one doesn't have to change, this is just informational for you.

Nasa62 (changes requested):

libzhlgen/test/functions/ELF_amd64/1.6.13/SoundControl.zhl
```diff
@@ -4,6 +4,8 @@ cleanup __amd64 void SoundControl::StopPlaylist(SoundControl *this, int fadeOut)
 cleanup __amd64 void SoundControl::StartPlaylist(SoundControl *this, std::vector<std::string> &playlist);
 ".415731c04156415541544989fc55534883ec48803f00":
 cleanup __amd64 int SoundControl::PlaySoundMix(SoundControl *this, const std::string& soundName, float volume, bool loop);
+"??57??56??55??5455??89fd53??83ec??803f????8974????f30f1104??74??":
```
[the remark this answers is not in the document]
ChronoVortex: Fixed. I still don't completely understand how `.` works, I'll have to ask you about it again at some point.

libzhlgen/test/functions/ELF_amd64/1.6.13/HackingSystem.zhl
ChronoVortex: Fixed

Nasa62 (approved):

PR #85, opened 2023-03-01
Added missing ln call in Linux docu

daniel86:


Nasa62:
This is not supposed to be a symlink to `lib` it's supposed to be a folder and then some more folders inside where you finally symlink ***just*** gcc-4.8.

If you symlink `lib` directly Clang will see all installed gcc versions and will use a newer one and the output library will not be compatible with FTL.

If your system literally only has GCC 4.8.4 or GCC 4.8.5 then you could get away with this, but you also wouldn't technically need the clang toolchain folder at all aside from the fact it's pre-defined in the codeblocks build.

PR #86, opened 2023-03-06
Add lowHealthThreshold crew XML tag

codyfun:
The HP amount for beeping and flashing can now be altered or the feature even disabled on a per-race basis. This required rewrites of CrewMember::DirectModifyHealth and CrewMember::OnRenderHealth.

Awarding of ACH_ROCK_FIRE works but has not been exhaustively tested for all scenarios.

### DIFF
--- a/Constants.h
+++ b/Constants.h
@@ -1,4 +1,4 @@
 #pragma once
 
-static const unsigned int numCachedStats = 52;
+static const unsigned int numCachedStats = 53;
 static const unsigned int numStats = numCachedStats + 7;
--- a/libzhlgen/test/functions/ELF_amd64/1.6.13/CachedRect.zhl
+++ b/libzhlgen/test/functions/ELF_amd64/1.6.13/CachedRect.zhl
@@ -1,4 +1,4 @@
 "41544189d45589f5533977104889fb":
 cleanup __amd64 void CachedRect::SetPosition(CachedRect *this, int x, int y);
-"41544189d45589f5533977184889fb":
+".41544189d45589f5533977184889fb":
 cleanup __amd64 void CachedRect::SetSize(CachedRect *this, int w, int h);
--- a/libzhlgen/test/functions/ELF_amd64/1.6.13/CachedRect.zhl
+++ b/libzhlgen/test/functions/ELF_amd64/1.6.13/CachedRect.zhl
@@ -1,0 +1,4 @@
+"41544189d45589f5533977104889fb":
+cleanup __amd64 void CachedRect::SetPosition(CachedRect *this, int x, int y);
+"41544189d45589f5533977184889fb":
+cleanup __amd64 void CachedRect::SetSize(CachedRect *this, int w, int h);
--- a/CrewMember_Extend.h
+++ b/CrewMember_Extend.h
@@ -228,6 +228,7 @@ public:
     std::string transformRace = "";
 
     float lowHealthThreshold = 25.f;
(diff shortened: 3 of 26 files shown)


libzhlgen/test/functions/ELF_amd64/1.6.13/CachedRect.zhl
```diff
@@ -0,0 +1,4 @@
+"41544189d45589f5533977104889fb":
+cleanup __amd64 void CachedRect::SetPosition(CachedRect *this, int x, int y);
+"41544189d45589f5533977184889fb":
```
Nasa62, line 3: These are directly after one another in the binary file, so the second one should have `.` for optimization purposes.

`".41544189d45589f5533977184889fb":`

(I mostly care about this for 64-bit since it's the primary version used for MV, x86 is fine as long as it's a unique match)

Constants.h
```diff
@@ -1,4 +1,4 @@
 #pragma once
 
-static const unsigned int numCachedStats = 52;
+static const unsigned int numCachedStats = 53;
```
Nasa62, line 3: @mathchamp93 this doesn't affect saving does it?

libzhlgen/test/functions/ELF_amd64/1.6.13/CachedRect.zhl
codyfun: Done!

Constants.h
mathchamp93: I don't believe the stat cache gets saved at all (that might change in the future though depending on what approach I use to deal with potential issues related to stats being checked in the middle of the game loading).

Even if it did, changing the number of stats would be automatically accommodated (it would break saved runs from old versions but the actual saving code would have been coded to use numCachedStats).

What the cache is used for is that when the game loads the max health will be cached immediately so that the cached value is used for the remainder of the loading process. Hyperspace directly saves and loads the health and max health of crew.

Also in the long run I might even remove the cache; it was an optimization I added before I found the root cause of performance issues (the custom augment code was poorly optimized and some augments got checked per crewmember so ships with lots of crew and/or augments would eat up a lot of CPU). I'd test the performance impact first, though.

mathchamp93:
Question, does lowHealthThreshold need to be a boostable stat? If it just needs to be a fixed value per race then we could save performance (though to be honest it's probably a very small performance hit in the grand scheme of things) by making it non-boostable (it would just use the crew definition's stat directly rather than calling CalculateStat).

I guess in theory there could be scenarios where you might want to have abilities/statboosts affect it (for example if you CrewSpawn temporary crew of an otherwise normal race).

codyfun:
It should be boostable in order to be adjusted alongside max health changes.

Nasa62 (approved):
I'm good with this, @mathchamp93 are you good with it since you had a question?

mathchamp93:
Wouldn't it be better to simply have two stats, one for flat low health threshold and one for fractional low health threshold? Then if the modder specifies one or the other it will use that one and ignore the other, and if the modder specifies both then it uses whichever of the two is lower at the time given the actual crewmember's health. Then modders wouldn't need to double up the stat boosts every time they want to make a health stat boost.

codyfun:
I think the CrewSpawn temporary variants (a la Clone Cannon) is still a good use case for a stat boost, but I've added the fractional threshold to simplify Parasitical Dependencies.

PR #89, opened 2023-03-31
fixed leaks from LoadFile calls

ChronoVortex:


mathchamp93 (approved):
Reviewed the changes. I'm still not 100% sure whether delete [] will work on the memory allocated by a method we're hooking but it didn't crash immediately so hopefully it's working.

PR #90, opened 2023-04-01
exposed rapidxml

ChronoVortex:


mathchamp93:
I've pushed an update so that the XML string itself is managed within the C++ classes rather than being a pointer to a string managed by Lua. This way the string and the document are guaranteed to live and die together.

This also allows LoadFile and parse to be combined into a single call that directly assigns the LoadFile result to the document.

mathchamp93 (approved):
I've contributed and added a class around the xml_document class to manage its own XML string (separate from the memory pool so that a string loaded from FTL can be assigned directly rather than having to do an extra copy). This should be a bit more robust since the string and document are guaranteed to have the same lifetime now, rather than the string potentially being garbage collected while the document is still being used (which could happen if the string is local/temporary but the document is kept as a global variable, for example). Also added convenience methods to parse a file in a single call or even construct and parse a file with just the constructor.

Note that other caveats related to memory management still apply. If you work with nodes/attributes and keep references to them, you must also keep a reference to the document itself, as once the document gets garbage collected the nodes/attributes are invalidated. Same goes if you parse a different file with the same document, clear the document, etc.

PR #91, opened 2023-04-07
Implemented .extend and .table members of relevant userdata

ChronoVortex:
Userdata types with corresponding Userdata_Extend classes can now access the extended class through the .extend member. These types also have a .table member, which accesses a unique table in the registry with a lifetime that corresponds to that of the userdata.


Nasa62:
Can you provide an example of how this is used in Lua for documentation?

Also does this break any backwards compatible use if someone used the previous Lua methods to access these classes?

Vertaalfout:
The old methods also work, and there's no backwards compatibility issue. Here's a simple example with projectiles.
[example.txt](https://github.com/FTL-Hyperspace/FTL-Hyperspace/files/11240293/example.txt)

Nasa62:
Added vert's example.txt as inline code so we can all see it easily
```lua
--TABLES: Certain types of userdata have .table member, which can be accessed and interacted with just like any other lua table. The contents of these tables are not saved.


local UNIQUE_TABLE_KEY = {} --Use a unique key to access our own table to avoid conflicts with other mods.

script.on_internal_event(Defines.InternalEvents.PROJECTILE_UPDATE_PRE, 
function(projectile)
  if not projectile.table[UNIQUE_TABLE_KEY] then --If our table was not initialized, initialize it and other members.
    projectile.table[UNIQUE_TABLE_KEY] = {timer = 0} 
  end 
  projectile.table[UNIQUE_TABLE_KEY].timer = projectile.table[UNIQUE_TABLE_KEY].timer + Hyperspace.FPS.SpeedFactor / 16
end, 2147483647)


script.on_game_event("CLEAR_ALL_RECENTLY_FIRED_PROJECTILES", false, 
function()
  local projectileList = Hyperspace.Global.GetInstance():GetCApp().world.space.projectiles
  for i = 0, projectileList:size() do
    local projectile = projectileList[i]
    if projectile.table[UNIQUE_TABLE_KEY].timer < 1 then
      projectile:Kill()
    end
  end
end)

--EXTEND: Certain types of userdata have a .extend member, which can be used to access a correlated class.

--The following two expressions are equivalent
print(SomeProjectile.extend.name) 
print(Hyperspace.Get_Projectile_Extend(SomeProjectile).name)
```

Nasa62:
@mathchamp93 how do you feel about this extend style, I know we ideally wanted something like `SomeProjectile.name` to just be the extended form and not `SomeProjectile.extend.name` but I guess it works.

Nasa62 (approved):
Add stuff to wiki before/when merging please.

Since this has merge commits in it, either rebase first & then to a regular merge or squash merge please.

PR #109, opened 2023-05-11
Exposed the game speed toggle boolean to lua.

Vertaalfout:
Added Hyperspace.FPS.speedEnabled as valid member in lua.


Nasa62:
It should target `develop`

PR #110, opened 2023-05-11
Replaced asserts in user-callable functions with luaL_argcheck

Vertaalfout:
Now improper arguments to lua functions will result in easy-to-understand lua errors, instead of assertion fails.


Nasa62:
I like it but lets bring this into `develop` so this can all go in with 1.4.0 in the future as it's not a bugfix

PR #112, opened 2023-05-12
Replaced asserts in user-callable functions with luaL_argcheck

Vertaalfout:
Now improper arguments to lua functions will result in easy-to-understand lua errors, instead of assertion fails.

### DIFF
--- a/lua/LuaLibScript.cpp
+++ b/lua/LuaLibScript.cpp
@@ -76,7 +76,7 @@ void LuaLibScript::LoadTypeInfo()
 
 int LuaLibScript::l_on_load(lua_State* lua)
 {
-    assert(lua_isfunction(lua, 1));
+    luaL_argcheck(lua, lua_isfunction(lua, 1), 1, "function expected!");
     LuaFunctionRef callbackReference = luaL_ref(lua, LUA_REGISTRYINDEX);
     m_on_load_callbacks.push_back(callbackReference);
     return 0;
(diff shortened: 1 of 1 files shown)


Nasa62 (changes requested):
My only concern here is that `luaL_argcheck` might not abort the execution?

`assert` I think also stopped the C/C++ function from continuing it's execution (like issuing a `return`)

Can we validate that it actually stops execution of the function? and/or have both or detect the lua error and stop execution?

ChronoVortex:
> Can we validate that it actually stops execution of the function? and/or have both or detect the lua error and stop execution?

`luaL_argcheck` aborted execution of the LUA script when I tested.

Nasa62:
> `luaL_argcheck` aborted execution of the LUA script when I tested.

Yes but I'm talking about the C-execution of that function where we're still doing other things

ChronoVortex:
> Yes but I'm talking about the C-execution of that function where we're still doing other things

Doesn't an error being raised stop execution of the function in C?

Nasa62:
> > Yes but I'm talking about the C-execution of that function where we're still doing other things
> 
> Doesn't an error being raised stop execution of the function in C?

I don't know, if it does, great. Does it?

ChronoVortex:
> I don't know, if it does, great. Does it?

```c++
#include <iostream>
#include <string>

void test() {
    throw 0;
    std::cout << "Hello, world" << std::endl;
}

int main() {
    try {
        test();
    } catch (int e) {
        std::cout << "An exception occurred. Exception #" << e << std::endl;
    }
}
```
```
An exception occurred. Exception #0
```
Yes.

Nasa62 (approved):
Cool then this is good

PR #123, opened 2023-07-01
Integrating lua into build & toolchains

nedsociety:
https://discord.com/channels/604415384979898464/701801787002323054/1121536752058892470

Some notes:
- This change involves globbing over `libzhlgen/test/functions` every time it builds. Unfortunately it's slow regardless of actual changes present or not. In WSL2-backed docker with host mounted repo it adds around ~8 seconds per build. Docker on Linux should take less since they don't involve the filesystem overhead.
- `FTLGame*.h` files are now generated files (in POV of CMake). I'd strongly recommend to remove and gitignore them from the repo, but for this PR I didn't since I can't verify CodeBlocks build is okay with it.

### DIFF
--- a/.devcontainer/Dockerfile-actual
+++ b/.devcontainer/Dockerfile-actual
@@ -9,8 +9,7 @@ COPY llvm.patch /llvm-build/
 RUN sed -Ei 's/^# deb-src /deb-src /' /etc/apt/sources.list \
     && apt update \
     && apt install -y --no-install-recommends \
-        ca-certificates gnupg build-essential dpkg-dev devscripts locales \
-        make python3 zlib1g wget subversion unzip git curl \
+        ca-certificates gnupg build-essential dpkg-dev devscripts locales make python3 zlib1g wget unzip git curl \
     && ln -s /usr/bin/autoconf /usr/bin/autoconf2.69 \
     && ln -s /usr/bin/autom4te /usr/bin/autom4te2.69 \
     && (LC_ALL=en_US.UTF-8 LANGUAGE=en_US.UTF-8 locale-gen en_US.UTF-8) \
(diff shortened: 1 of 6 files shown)


Nasa62:
> * `FTLGame*.h` files are now generated files (in POV of CMake). I'd strongly recommend to remove and gitignore them from the repo, but for this PR I didn't since I can't verify CodeBlocks build is okay with it.

Yes we should please have them deleted & `.gitignore`'d please.
You can modify the cbp project file to just remove the `FTLGame*` files and it should still work because they'll be picked up in the actual build I think as includes just like how some files like `Version.autogen.hpp` are not defined there but still get pulled in (they just don't show up in the IDE for editing)

Unfortunately I don't think they can show up as `auto-generated` like the SWIG .i files do (creating .cpp files under auto-generated) since that requires a different file extension and it has to be setup in the CodeBlocks compiler advanced settings (which is not part of the project file and not synchronized between developers computers and is annoying to do anything but leave as defaults)

nedsociety:
[710b46c](https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/123/commits/710b46ca17fde3d0401cbc9b1dd7609153639a94)

nedsociety:
```
/workspaces/FTL-Hyperspace/lua/modules/hyperspace.i:2848: Error: Unable to find 'FTLGameELF64.h'
```
(TODO: this is unintended and should be fixed -- FTLGame*.h should be generated before the SWIG step)

Nasa62:
> [710b46c](https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/123/commits/710b46ca17fde3d0401cbc9b1dd7609153639a94)
> 
> > You can modify the cbp project file to just remove the FTLGame* files and it should still work because they'll be picked up in the actual build
> 
> Note that cbp does not have an instruction to run zhlgen so it can't generate those files. The first build right after the clean clone would fail on CodeBlocks if one didn't run zhlgen manually beforehand.

I believe you can set it to run the script before build just like I did for the version autogen script

nedsociety:
> I believe you can set it to run the script before build just like I did for the version autogen script

Done, I can't promise it works though.

Nasa62:
@nedsociety was this the one that needed an SSH key? If not, can we resolve the conflicts so we can get this one in?

nedsociety:
It's not ssh-related and that can be done separately.

nedsociety:
@Nasa62 rebased to the latest master. This comment still holds:

> The devcontainer image must be rebuilt for it to work (CI would fail for this PR).

PR #124, opened 2023-07-01
Deprecate develop in CI

nedsociety:


.github/workflows/release.yml
```diff
@@ -58,7 +42,7 @@ jobs:
 
       - name: Push changes to the repo
         run: |
-          git push origin develop:develop master:master
+          git push origin master:master
```
Nasa62, line 45: Would it be possible for it to make a "Release PR" that has the version update? So we can do that to choose if we're actually releasing a new version or if we need to tweak the version to a major/minor etc...?

And then it can do the release without updating the version on that next run?

Nasa62 (approved):

.github/workflows/release.yml
nedsociety: That would run _two_ workflows per release -- one for version bump, and one for release. Wouldn't that be excessively complicated?

Nasa62: Maybe? I mean the first step is it creates the PR to bump version, then if and only if the commit message matches that PR (or maybe it detects there was a change to the version in the last commit) that it runs the tag & release workflow?

I don't think it's that complicated but I'm open to other suggestions of giving us a way to control the version numbers.

Nasa62: Should be fairly easy to detect "did a version number change in the last release" by just checking if there was a change to HSVersion.h in the details of the latest commit (or if you know the last released commit, you could do it over a range)

Nasa62: Or actually I see, are we supposed to do the release by running the workflow manually and setting the version number there?

If so then that's just fine and I'll get you a SSH-key for the automation.

nedsociety: > Or actually I see, are we supposed to do the release by running the workflow manually and setting the version number there?

Yup this should be the case.

PR #135, opened 2023-07-15
Updated German Translation

The-Dumb-Dino:
I Updated the german translation to HS-1.6. As a little sidenote; it took me 2 hours to find all the new texts because I had no way to compare them. Please add a comment to new texts so that I can see whats missing the next time.


Nasa62:
@ChronoVortex you want to glance at this too? It looks okay but I didn't test it.

PR #141, opened 2023-07-19
Added ability to prevent achievements from showing for specific ships

Vertaalfout:


ChronoVortex (changes requested):
Something is very wrong, my game crashes as soon as I hit "New Game"

ChronoVortex (approved):

Issue #134, opened 2023-07-13
Fix the crash that occurs when the ship doesn't have a shield system assigned and a overcharger drone trigger.


gabrielsubzero:
Currently if you don't have a shield system assigned to a ship and a shield drone triggers the game will crash, not only that, not having shields also makes energy shields not work at all, i assume it's a vanilla bug but it's very annoying to work around imo.

Nasa62:
This is probably because the ShieldSystem doesn't actually exist in that case, so it's calling a null value... are Shield drones a Hyperspace feature?

ChronoVortex:
Fixed in https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/147

PR #147, opened 2023-07-24
Rewrote SuperShieldDrone::OnLoop to prevent crash in vanilla (See issue #134)

Vertaalfout:
Before this rewrite, overcharger drones would crash when attempting to add a supershield to a ship with a null shieldSystem (resulting from having <shields> undefined in the shipBlueprint).


ChronoVortex (changes requested):
This looked good in testing, but I'd like you to fix your tabs before I merge, they use 2 spaces instead of 4 which is inconsistent with the rest of the codebase

ChronoVortex (dismissed):

ChronoVortex (approved):

PR #148, opened 2023-07-25
Fix: German Translation

The-Dumb-Dino:
Fixed an UI Bug in the german translation. For some reason a text in the hangar was shifted to the left (Caused by Hyperspace).
I came up with the janky and primitive solution to embed it with its UI onto the hangar image. It is janky but works great and can not be seen by the player. (I didnt change the actual hangar image, it will only be present when the games language is set to german.


The-Dumb-Dino:
image

The-Dumb-Dino:
image
Thats how it looks ingame, you can not see it at all

ChronoVortex (approved):

Issue #114, opened 2023-05-27
'LUA print("Hello World")' example in hyperspace.xml does not work on non-Linux or non-dev version of Windows


CommunistMountain:
Change it to 'LUA log("Hello World")'

Nasa62:
Closed, it remains `print` because of the new print implementation in #149

Issue #128, opened 2023-07-06
deaf accessibility


Rattman1233:
ftl doesnt have that many audio-only cues and most audio cues have something visual to accompany them, but playing the game audioless is still pretty hard (try it yourself if you dont believe me). hyperspace should provide functionality to create an accessibility mod similar to minecrafts closed captions, bascially being able to create a list/log of all sounds played in an easy to reference place.

Nasa62:
#149 should make this more practical now, we just need to add a hook into the actual audio playing functions and log the names of the sounds being made (or hook it into Lua and then someone can write a translated deaf-accessibility thing there, and could also pop up icons or other graphics for certain sounds if they want)

PR #149, opened 2023-07-26
New print function for lua

Vertaalfout:
`print` function in lua now writes messages to your screen. The location, font, and other relevant details can be configured using the `Hyperspace` module in lua.

Usage:
```lua
print("Hello, world!")
print(1,2,3,4,5)
print(Hyperspace.ships.player)
```

Configuration:
```lua
Hyperspace.PrintHelper.x = 300
Hyperspace.PrintHelper.useSpeed = true
--Alternatively
Hyperspace.PrintHelper.GetInstance().x = 300
Hyperspace.PrintHelper.GetInstance().useSpeed = true
```


lua/luaGlobal.h
```diff
@@ -41,11 +42,37 @@ static int hslua_g_printMessage(lua_State* lua)
...
+    printf("%s\n", message.c_str());
+    return 0;
+}
+
 // TODO: Maybe restructure this as a class and have this be part of constructor?
 // TODO: Or maybe constructor calls the global singleton of LuaScriptInit instance and gets the lua state?
 void hsluaglobal_register(lua_State* lua)
 {
     lua_register(lua, "log", hslua_g_printMessage);
+    lua_register(lua, "print", hslua_g_printToScreen);
```
Nasa62, line 75: `print` already exists as a word if I remember right.
This is technically overwriting the existing `print` method from Lua that writes to the standard out.

Technically we probably should make it write to the hs log (and standard out) like log already does and like we probably should have made print do in the first place.

lua/luaGlobal.h
```diff
@@ -41,11 +42,37 @@ static int hslua_g_printMessage(lua_State* lua)
...
+        {
+            return luaL_error(lua, "'tostring' must return a string to 'print'");
+        }
+        
+        message += str;
+        message += "    ";
+        lua_pop(lua, 1);
+    }
+    PrintHelper::GetInstance()->AddMessage(message);
+    printf("%s\n", message.c_str());
```
Nasa62, line 66: Lets also call
```c++
hs_log_file("%s", logMessage.c_str());
```
So that it's logged to both.
Lets also please format the message with `[Lua]: %s\n` like the log message does so it's clear which logs are from lua and which are from other HS internals.

Nasa62 (changes requested):

Nasa62:
Will resolve #114 if merged

lua/luaGlobal.h
Vertaalfout: Should I just do that for `printf` and `hs_log_file` because those will be used by HS internals? Orr should I do it for the message used by `PrintHelper` as well? I'm inclined to do it for all three  because it is possible that `PrintHelper` would be used by some HS internals in the future, such as issue #128.

Nasa62: Well `PrintHelper` displays on screen right?
Humm... #128 is a good point though...

Actually no, don't do it for `PrintHelper` because this is used to show messages to the user, the other is really more for us having a log for debugging and knowing "oh those lines came from Lua, etc..." and not from something else within HS.

If they're logging a error-case they can just use `log(` to prevent it from displaying on-screen to the user.

Vertaalfout: Made that change, also added varargs/non-string arg support for `log`

Nasa62 (approved):

PR #152, opened 2023-07-26
Added __call field to ships metatable

Vertaalfout:
Added __call field to ships metatable.

Usage:
```lua
Hyperspace.ships(0) --Equivalent to Hyperspace.ships.player
Hyperspace.ships(1) --Equivalent to Hyperspace.ships.enemy

Hyperspace.ships(shipId) --If invalid shipId, throws error, otherwise equivalent to:
Hyperspace.Global.GetInstance():GetShipManager(shipId)
```


Nasa62 (approved):
Yeah I like this

Issue #132, opened 2023-07-08
Fix the crash caused by when you deal 10 or more ion damage in a single hit to a un-ionized system while having a manned sensors 3.


gabrielsubzero:
Basically the game crashes rn if you can see a ion stack that's 10 or higher, it would be nice if it was fixed, preferably in a way that also shows ion stacks higher than 5, as rn there isn't a number for those so it's just blank iirc.

Nasa62:
I do believe this is vanilla, as the game only has an array of images for 9 ion states, and it's hardcoded. It would probably require completely reverse engineering how ion rendering is handled in FTL.

Nasa62:
Will be fixed by #161 when merged

PR #161, opened 2023-07-30
Fixed ion bug (Issue #132)

Vertaalfout:
Fixed bug where when a system had more than 9 ion damage applied, it crashed the game due to rendering an invalid primitive off the end of the timerCircle array


libzhlgen/test/functions/ELF_amd64/1.6.13/ShipSystem.zhl
```diff
@@ -98,6 +98,12 @@ cleanup __amd64 void ShipSystem::constructor(ShipSystem *this, int systemId, int
 "55534889fbbf????????4883ec18f30f1144240c":
 cleanup __amd64 bool ShipSystem::PartialDamage(ShipSystem *this, float amount); /* Thunk of ShipSystem::DamageOverTime, maybe it can be used instead? */
 
+"4157b840":
```
Nasa62, line 101: These new signatures should ideally be relative to speed up the 64-bit scan.
Also have not validated yet.

Misc.cpp
```diff
@@ -99,3 +99,223 @@ HOOK_METHOD(SpaceManager, SwitchBackground, (const std::string& backgroundList)
...
     return ret;
 }
+
+//TODO: Move to appropriate file and cleanup a bit
+static GL_Primitive* timerCircleMax = nullptr;
+static bool initExtraPrimitive = false;
+
+HOOK_METHOD(SystemBox, constructor, (Point pos, ShipSystem *sys, bool playerUI) -> void)
+{
+    if (!initExtraPrimitive)
```
Nasa62, line 109: Hook trace logs are missing from this file, run the script to generate them

ChronoVortex (approved):

PR #163, opened 2023-07-30
Implemented noWarning statboost

Vertaalfout:
Implemented statboost to prevent crew from counting towards the warning when you attempt to leave without recalling them from the enemy ship.


libzhlgen/test/functions/ELF_amd64/1.6.13/ShipManager.zhl
```diff
@@ -182,7 +182,8 @@ cleanup __amd64 int ShipManager::constructor(ShipManager *this, int shipId);
 cleanup __amd64 void ShipManager::destructor(ShipManager *this);
 "!.534889fb":
 cleanup __amd64 void ShipManager::destructor2(ShipManager *this);
-
+"41564989fe415541545553488b87c800":
```
Nasa62, line 185: This signature should be adjusted to be relative to the other signatures and not absolute as per the standard for all my other 64-bit signatures when possible

ChronoVortex (approved):

PR #175, opened 2023-08-09
Added SKILL command

Vertaalfout:
New SKILL command to fully level up all of your crew.


Mwallx:
If it's possible I'd like 2 arguments so that we no longer need to waste our time to train the crew when the enemy is harmless. So maybe something like SKILL 2 PILOT then it sets the second crew in the left list to max in pilot skill.

Vertaalfout:
This is mostly because I know a fair few players who like to farm skills, and this is mostly intended for that. I think that it would probably be better to just expose skill levels to lua (if they are not already exposed) if people have a need for setting specific skill levels for some form of testing.

Nasa62:
> This is mostly because I know a fair few players who like to farm skills, and this is mostly intended for that. I think that it would probably be better to just expose skill levels to lua (if they are not already exposed) if people have a need for setting specific skill levels for some form of testing.

True, maybe we could add setting skills of a specific crew as an example to the wiki for running with the `LUA` command in the console.

ChronoVortex (approved):

PR #176, opened 2023-08-09
add workflow to update wiki from master branch

ChronoVortex:
Changes made to files in the `wiki` folder will be automatically applied to the wiki when pushed to `master`.

`updateWiki.sh` taken from https://github.com/orlovm/Wiki-Action


updateWiki.sh
Nasa62, line 1: Maybe we could move this under `.github/workflows` or something? So it's not polluting the root directory?

wiki/Finding-Signatures-with-Ghidra.md
```diff
@@ -0,0 +1,207 @@
...
+         - **From Linux/Mac/Windows**: Download the `1.6.12` zip directly, unzip somewhere
+      - **Mac Version**
+         - **From Linux/Mac/Windows**: Download the `.dmg` directly, I think Ghidra can open this directly (but untested)
+   - **Epic/Microsoft/Other Sources**
+      No Linux or Mac binaries provided, unknown how to get old Windows binaries other than the downgraders on the forums
+3. Copy all the binaries you wish to examine to their own project folder for Ghidra, you don't need to keep them in their original locations
+4. Open them in Ghidra, see special instructions for each type:
+   - Windows Binaries
+      - After importing the EXE but before opening it you must change the language
+      - ![image](https://user-images.githubusercontent.com/1423894/172201781-e00e2107-4feb-47d3-bd45-2938811b5fcc.png)
```
Nasa62, line 42: Obviously this works for now because it was manually uploaded in the past, but maybe we should also test if we have a `wiki/img` folder or something and if relative links work?

Nasa62 (dismissed):

Nasa62:
Looks great, great work! This is super awesome and lets us update the wiki and add new features in the same PR.

updateWiki.sh
ChronoVortex: Done

wiki/Finding-Signatures-with-Ghidra.md
ChronoVortex: That does in fact work, implemented

Nasa62 (approved):
Did you test if the relative image files worked?

ChronoVortex:
> Did you test if the relative image files worked?

Yep, made a temporary repo to use as a sandbox

PR #179, opened 2023-08-14
Implemented ON_WAIT callback

Vertaalfout:


ChronoVortex (changes requested):
Should pass the `ShipManager` to the callback like the other jump events.

Vertaalfout:
This is only ever called on the player's ship, so I originally wanted this to be reflective of that, but out of all the jump callbacks only `JumpLeave` is ever called on the enemy ship, so internal consistency is probably better regardless

ChronoVortex (approved):

PR #181, opened 2023-08-17
Button and Mouse exposure and new callback

Vertaalfout:
Exposed GenericButton, Button, MouseControl, added proper constructors for Global::Ellipse and Global::Rect, exposed TimerHelper::Start and float overload.

NOTE: GUI_MOUSE_MOVE was implemented because ON_MOUSE_MOVE doesn't cause the button mouseover beep and mouse cursor animation change, this prevents a need for jank implementations in lua.


lua/InternalEvents.h
ChronoVortex, line 1: This should be added to the Defines module wiki page.

lua/modules/hyperspace.i
ChronoVortex, line 1: Exposed classes and their members should be added to the Hyperspace Module wiki page.

ChronoVortex (changes requested):

ChronoVortex (approved):

PR #183, opened 2023-08-18
Changed image arg so string conversion works properly in lua

Vertaalfout:


Nasa62 (approved):
@ChronoVortex it looks good but I didn't test it.

PR #221, opened 2024-02-05
Adding Cooldown Callback 

THETASOLA:
Adds another lua callback, centered on WeaponBox::RenderBox, dubbed WEAPON_RENDERBOX.

`function weapon_renderbox(ProjectileFactory& weapon, int cooldown, int maxCooldown, char* firstLine, char* secondLine) return Chain, secondLine, firstLine`

The main purpose of this callback is to dynamically overwrite a weapon cooldown status on top of the WeaponBox, there is two line, the first being the weapon cooldown/max cooldown. The second being previously used exclusively by adaptive weapon (increases damage with each shot with a visual "x damage").

image


THETASOLA:
Did not make enough verification, second line is incorrect.

PR #223, opened 2024-02-06
Expose AddEquipment and RemoveEquipment

Choosechee:
Expose AddEquipment and RemoveEquipment so we can give blue options without having the item normally required.


wiki/Lua-Hyperspace-Module.md
ChronoVortex, line 1: This is the only file that should be changed in this PR, everything else needs to be brought back in line with FTL-Hyperspace:master.

ChronoVortex (changes requested):

PR #224, opened 2024-02-07
Exposing CommandGUI:RunCommand and CommandConsole

arcburnergit:
This allows us to do make use of some of the features of the console within lua, such as loading events with ships after already having loaded a ship.

### DIFF
--- a/wiki/Lua-Hyperspace-Module.md
+++ b/wiki/Lua-Hyperspace-Module.md
@@ -1090,23 +1090,23 @@ The members held by this class determine how commands run in the console behave.
    - Runs a vanilla console command.
 
 ### Fields
-- `ShipStatus` `shipStatus`
-- `CrewControl` `crewControl`
-- `SystemControl` `sysControl`
-- `CombatControl` `combatControl`
-- `FTLButton` `ftlButton`
-- `SpaceStatus` `spaceStatus`
-- `Point` `pauseTextLoc`
-- `Point` `shipPosition`
-- `bool` `outOfFuel`
-- `bool` `bPaused`
-- `bool` `bAutoPaused`
-- `bool` `menu_pause`
-- `bool` `event_pause`
-- [`Button`](#button) `upgradeButton`
-- `bool` `dangerLocation`
-- `bool` `bHideUI`
-- `bool` `jumpComplete`
-- `int` `mapId`
-- `bool` `secretSector`
-- `bool` `choiceBoxOpen`
+- `ShipStatus` `.shipStatus`
+- `CrewControl` `.crewControl`
+- `SystemControl` `.sysControl`
+- `CombatControl` `.combatControl`
+- `FTLButton` `.ftlButton`
+- `SpaceStatus` `.spaceStatus`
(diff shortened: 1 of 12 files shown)


wiki/Lua-Hyperspace-Module.md
```diff
@@ -1069,3 +1069,44 @@ local _, canMove = crew.extend:CalculateStat(Hyperspace.CrewStat.CAN_MOVE)
...
+
+### Methods
+- `bool :RunCommand(CommandGui *commandGui, const std::string& command)`
+   - Used for running a Hyperspace command.
+- `CommandConsole .GetInstance()`
+   - Returns the main instance of `CommandConsole`. Always use this to access any members belonging to this class.
+
+### Fields
+- `bool` `.enabled`
+   - Whether or not the Command Console is enabled or not, this is set in Hyperspace.xml.
```
ChronoVortex, line 1084: You have a redundant "or not" in this sentence, one of them needs to be removed.

wiki/Lua-Hyperspace-Module.md
```diff
@@ -1069,3 +1069,44 @@ local _, canMove = crew.extend:CalculateStat(Hyperspace.CrewStat.CAN_MOVE)
...
+
+## CommandGUI
```
ChronoVortex, line 1086: The case of this struct's name is `CommandGui`, not `CommandGUI`.

wiki/Lua-Hyperspace-Module.md
```diff
@@ -1069,3 +1069,44 @@ local _, canMove = crew.extend:CalculateStat(Hyperspace.CrewStat.CAN_MOVE)
...
+- `bool bAutoPaused`
+- `bool menu_pause`
+- `bool event_pause`
+- `Button upgradeButton`
+- `bool dangerLocation`
+- `bool bHideUI`
+- `bool jumpComplete`
+- `int mapId`
+- `bool secretSector`
+- `bool choiceBoxOpen`
```
ChronoVortex, line 1112: This doesn't follow the formatting of fields in the rest of the page. for example, `ShipStatus shipStatus` should be `ShipStatus` `.shipStatus`. If any of the classes of these fields are exposed and documented, there should also be links leading to their section in this page.

ChronoVortex (changes requested):

Nasa62:
@ChronoVortex I'm not sure how I feel about this.
I'd almost rather implement the console features in Lua and instead change the command console to a Lua input but with some backwards compatibility for the old commands rather than allowing Lua to hook into the console.

Kinda a catch-22, if we allow this, then Lua code can be forever tied to console commands, instead of exposing the actions those commands take to Lua and allowing us to clean up the console in the future so it's more of a proper scripting console with some hardcoded default commands like most other games would have first.

(so like the default input is lua, unless you use one of the special words that matches the old commands for compatibility)

I guess lua->console even when there's a "improved console" could still work but it feels weird... I feel like having those features accessible from Lua instead is better than running weird console commands via Lua.

PR #225, opened 2024-02-07
Locked Miniship Icon prioritize _base.png version

THETASOLA:
Add a detection and a replacement of the locked miniship image by their _base variant, notably used by the discord integration.
image
This allows users to be able to better see the shape of the locked ship, with no icon "masking" the part of the hull.

image
(no replacement when no _base miniship exists)


CustomShipSelect.cpp
```diff
@@ -909,6 +909,15 @@ void CustomShipSelect::OnInit(ShipSelect* shipSelect_)
                 {
                     aButton->bShipLocked = !CustomShipUnlocks::instance->GetCustomShipUnlocked(shipName);
                     aButton->bLayoutLocked = aButton->bShipLocked;
+
+                    if (aButton->bShipLocked){
```
ChronoVortex, line 913: Brackets should respect the formatting of existing code. Change from
```c++
if (aButton->bShipLocked){
```
to
```c++
if (aButton->bShipLocked)
{
```
for all instances.

ChronoVortex (changes requested):

ChronoVortex (approved):

PR #226, opened 2024-02-07
Adding Cooldown Callback

THETASOLA:
Adds another lua callback, centered on WeaponBox::RenderBox, dubbed WEAPON_RENDERBOX.

function weapon_renderbox(ProjectileFactory& weapon, int cooldown, int maxCooldown, char* firstLine, char* secondLine) return Chain, secondLine, firstLine

The main purpose of this callback is to dynamically overwrite a weapon cooldown status on top of the WeaponBox, there is two line, the first being the weapon cooldown/max cooldown. The second being previously used exclusively by adaptive weapon (increases damage with each shot with a visual "x damage").

image


ChronoVortex (changes requested):
This works fine when the callback doesn't return anything, but when it does I'm getting some bugs with the LUA stack that I can't figure out. Using this code
```lua
script.on_internal_event(Defines.InternalEvents.WEAPON_RENDERBOX, function(weapon, cooldown, maxCooldown, firstLine, secondLine)
    return "hello"..firstLine, "world"..secondLine
end)
```
for a chain weapon I get the result
```
1 Damage
world1 Damage
```
as opposed to the expected
```
world1 Damage
hello0.0/10.0
```
which suggests that `streamStr = lua_tostring(context->GetLua(), -2)` is getting the second return value instead of the first and `stream2Str = lua_tostring(context->GetLua(), -1)` is getting nothing because the value at offset -1 isn't even a string. No idea why this is happening, this callback will have to wait until we can figure it out.

ChronoVortex (approved):
Turns out it works fine, just forgot `Defines.Chain.Continue`

PR #238, opened 2024-02-16
Refactoring tooltips

Admiral-Billy:
Stat appearances may not match the previous version 1:1, so a wee bit of peeping would be appreciated, but from what I looked at they seemed to be the same.

Preemptive weapons use a new "max_shots" text library entry, e.g. most preemptive weapons will say something like "Max shots: 1".


ChronoVortex (approved):
A perfect PR that never had any issues :)

PR #239, opened 2024-02-16
mac build

ChronoVortex:
Re-opened after the original target branch was deleted. The last time this PR was updated was during 1.8.0, so all the signatures that were added since then need to be added to the Mac ZHL. Aside from that, remaining work will consist primarily of rewriting vanilla functions that were inlined on the Mac build of FTL but not on other builds.


The-Dumb-Dino:
May this never be forgotten

PR #251, opened 2024-03-05
Add a "force" attribute that allows recallBoarders to recall crew who can't teleport

Choosechee:
Right now, recallBoarders doesn't recall player crew who can't teleport. This is really bad for the way Multiverse uses it, since these crew will be lost unfairly. For some reason, a check for whether a crew can teleport is directly in the RecallBoarders method itself. I thought it was an oversight where it calls a method used to teleport crew, and that method checks if a crew can teleport, but it seems to have been intentional. I have added a "force" attribute that makes the check always pass, regardless of whether a crew can teleport.


ChronoVortex (commented):
Changes like this aren't great for backwards compatibility. Instead of removing the check entirely, you should add an attribute to the `recallBoarders` tag which determines whether all crew are recalled or just the ones that can teleport.

Choosechee:
Alright, I've made it an attribute.

CustomEvents.cpp
```diff
@@ -1059,6 +1059,10 @@ bool CustomEventsParser::ParseCustomEvent(rapidxml::xml_node<char> *node, Custom
                 else if (strcmp(stringValue, "player") == 0) {
                     customEvent->recallBoardersShip = -1;
                 }
+            }
+			if (child->first_attribute("force"))
```
ChronoVortex, line 1063: This line uses the wrong kind of indentation, should be four spaces instead of the tab character.

CustomEvents.h
```diff
@@ -621,6 +621,7 @@ struct CustomEvent
     std::string secretSectorWarp = "";
     bool recallBoarders = false;
     int recallBoardersShip = 0;
+	bool recallBoardersForce = false;
```
ChronoVortex, line 624: Wrong indentation here too.

ChronoVortex (changes requested):

ChronoVortex (approved):

PR #256, opened 2024-03-12
Complete the missing methods of `Ship` class.

kokoro11:
Windows version has been tested on my computer and runs without errors. Linux versions are not tested.

### DIFF
--- a/FTLGameELF32.cpp
+++ b/FTLGameELF32.cpp
@@ -12000,137 +12000,137 @@ namespace _var874
 namespace _func875
 {
 	static void *func = 0;
-	static short argdata[] = {0x1ff, 0x1ff, 0x1ff, 0x1ff};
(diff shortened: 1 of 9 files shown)


ChronoVortex:
What exactly is the purpose of exposing all these? We usually expose functions and members as needed, this would be a lot to test at once

kokoro11:
> What exactly is the purpose of exposing all these? We usually expose functions and members as needed, this would be a lot to test at once

Well I just thought it would be nice to make these functions exposed? ...if I must say, there are a lot of rendering functions in there that can help us achieve better control over rendering ship layers (e.g. `OnRenderWalls`, `OnRenderSystem`...).
But anyway, there are some bugs in there that should certainly be fixed, for instance, the windows version of `OnRenderSparks` actually hooks `OnRenderHighlight`. (the signature used for that function is `8b51088b410c29d0c1f80285c074??5589e5565389ce31db83e4f0908d7426008b0c9a83c301`, which matches 3 functions, and the first one is OnRenderHighlight instead of `OnRenderSparks`) `BreachSpecificHull` has a return type of `bool` instead of `void`. `EmptySlots` has a return type of `void` instead of `int`.

ChronoVortex (changes requested):
In that case, I'd prefer you apply those bug fixes and add the specific render functions that you want. I simply do not have time to validate every single one of these signatures and test all of them in both Windows and Linux.

PR #262, opened 2024-03-25
Make downgrade.bat work when run as administrator

compiling:
When a batch file is run as an administrator through the windows right click menu, it is started with the current working directory set to C:\Windows\System32.  To work around this, change directory to the path of the batch file at the top of the file.


Nasa62:
Is there a reason people are running it as administrator?

compiling:
You need to run it as an admin if FTL is installed in Program Files and you have the standard windows access rules. I believe that is actually the default location for Steam.

Nasa62 (approved):

PR #263, opened 2024-03-26
Expose ShipInfo::augList

Choosechee:
There's no way to check the hidden augments of a ship in Lua right now. Exposing CustomAugmentManager would provide the method GetShipAugments, which would hopefully allow you to check them. IsAugment could also be a useful method for adding cross compatibility with addons.


Nasa62:
Please add the corresponding documentation entry in the wiki files (wiki folder)

Choosechee:
> Please add the corresponding documentation entry in the wiki files (wiki folder)

Done.

Nasa62 (dismissed):

Choosechee:
I noticed this line in hyperspace.i:
`%template(unordered_map_string_int) unordered_map<string,int>; // line 126`
(I added the comment with the line number). Shouldn't this template be the one that hiddenAugs needs? Why didn't it work when I tested it? Is hiddenAugs an ordered map?

ChronoVortex (changes requested):
Checking the code, it doesn't look like `ShipManager_Extend::hiddenAugs` is even referenced anywhere after it's declared, so it's probably not what you want. Try `CustomAugmentManager::GetShipAugments`, that one might return a list that includes the hidden augments.

ChronoVortex (approved):
It turns out that it was impossible to distinguish between hidden augments and normal augments in the list returned by `GetShipAugments`. The only place I could find where they are distinguishable is `ShipInfo::augList`, where hidden augments have the `HIDDEN` prefix. You can get all the augments for the player ship with this code:
```lua
local augs = Hyperspace.Global.GetInstance():GetShipInfo(false).augList:keys()
```
`augs` will be a `vector` containing all the augments on the ship. Using this on the MV Kestrel at the start of the game, for example, gives this list:
```
FTL_JUMPER_SCANNER    
HIDDEN FOR_MULTIVERSE    
HIDDEN MIND_ORDER    
HIDDEN SHIP_MULTIVERSE    
```

PR #267, opened 2024-04-01
added SELinux exception instructions

### LINKED ISSUE #195: Can't run the `FTL` script on linux after following the guide: `Fatal Error Failed to install hook for WeaponControl::RenderAiming: Failed to make copy of original code executable:`
I'm trying to install FTL Multiverse on my GNU/Linux system (Fedora 38 with KDE Plasma). I'm working with this (well written and officially seeming) [guide](https://ftl-hyperspace.github.io/FTL-Hyperspace/install-guides/linux/genericlinux/steam-install). I get a fatal error sadly.

**Setup**

I have inserted the 3 shared objects:
```
Hyperspace.1.6.12.amd64.so 
Hyperspace.1.6.13.amd64.so 
liblua5.3.so.0             
```
As documented. I have also included the 2 extra exports in the `FTL` shell script:
```bash
#!/bin/sh


here="$(dirname "$(readlink -f "$(which "$0")")")"
there="$(pwd)"

command="$(basename "$0")"

machine="$(uname -m)"
if [ -e "$here/$command.amd64" ] \
   && ( [ "$machine" = 'x86_64' ] || [ "$machine" = 'amd64' ] )
	then arch='amd64'
	else arch='x86'
fi

echo Loading Arch = $arch

export LANG="${LC_ALL:-$LANG}"
export LC_ALL=
export LC_NUMERIC=C
export LD_LIBRARY_PATH="$here":${LD_LIBRARY_PATH} # extra for FTL Hyperspace
export LD_PRELOAD=Hyperspace.1.6.13.amd64.so # extra for FTL Hyperspace
exec "$here/$command.$arch" "$@"
```
(the comments are not part of the actual file and only added for this issue)

My FTL Folder in steam looks like this:
```
.
├── data
│   ├── amd64
│   │   └── libsteam_api.so
│   ├── exe_icon.bmp
│   ├── FTL
│   ├── FTL.amd64
│   ├── ftl.dat
│   ├── FTL.log
│   ├── FTL.x86
│   ├── Hyperspace.1.6.12.amd64.so
│   ├── Hyperspace.1.6.13.amd64.so
│   ├── liblua5.3.so.0
│   ├── licenses
│   │   ├── README-FreeType.txt
│   │   ├── README-RapidXML.txt
│   │   └── README-zlib.txt
│   ├── x86
│   │   └── libsteam_api.so
│   └── zhl.log
├── FTL
└── FTL_README.html

5 directories, 17 files
```

I have also patched hyperspace into FTL using Slipstream.

I did the setup process two times to be sure.

**Problem**
Now when running, some hook cannot be installed, resulting in a fatal error.
```
$ ./FTL              
Loading Arch = amd64
ELF Relocation addr: 0x0000000000000000
ELF Loaded at 0x0000000000400000, 0x0000000000400000
SEG SIZE: 0x0000000000200000
SEG SIZE: 0x0000000000200000
SEG SIZE: 0x0000000000200000
Fatal Error Failed to install hook for WeaponControl::RenderAiming: Failed to make copy of original code executable:                            
```

I assume, that this line causes the behaviour:
```
export LD_LIBRARY_PATH="$here":${LD_LIBRARY_PATH}
```
What happens here? It seems to me that `LD_LIBRARY_PATH` is set to itself, confusing the shell.

**Attempts at Troubleshooting**
When removing the line with `LD_LIBRARY_PATH`, the shell script starts the game, and according to the game version displayed in the corner, Hyperspace 1.8.0 is loaded according to that, but I assume that is just because of the Slipstream patch. In that case I also get a couple of warnings aswell as the error that the shared object could not be preloaded.
```
ERROR: ld.so: object 'Hyperspace.1.6.13.amd64.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
```
I assume the game just runs as vanilla at this point, since the actual libraries are not loaded. Steam can't start the game like this.

kali-muon:
To the troubleshooting section, I added instructions for how to fix SELinux preventing Hyperspace from launching. 

I wasn't able to get the test server running after poking around at it for a while, so I thought I would just submit this PR for now. If you require PRs to be tested, I can figure out how to get the test server working.

### DIFF
--- a/_i18n/en/install-guides/linux/index.md
+++ b/_i18n/en/install-guides/linux/index.md
@@ -35,3 +35,19 @@ To potentially fix a corrupted save:
 5. Move your original `hs_mv_prof.sav` back where it was before, replacing the one that was created by the game
 6. If this still doesn't work, delete the entire `~/.local/share/FasterThanLight/` folder
 7. ALso disable Steam Cloud Sync if you are using a version of FTL from steam, it'll frequently break the save files.
+
+#### Running FTL gives an error along the lines of "Failed to install hook"
+Launching FTL may give this error:  
+`Fatal Error Failed to install hook for WeaponControl::RenderAiming: Failed to make copy of original code executable:`  
+If you get this error, it's likely that SELinux is preventing the Hyperspace library from being linked. SELinux is common on Fedora Linux. To fix SELinux's interference:
+1. Verify that SELinux is running:  
+```$ sestatus```
+2. If it's enabled, verify that it's in Enforcing mode:  
+```$ getenforce```
+3. If it is in Enforcing mode, we need to add an exception in SELinux. We do that using `ausearch` to find the SELinux error message, `audit2allow` to generate the exception rule, and `semodule` to install the rule. (If your architecture is not `amd64`, change the `amd64` to the FTL executable your architecture runs.) If there are no search results for the error, you need to launch FTL again to generate the error so `ausearch` can grab it.  
+```$ sudo ausearch -c "FTL.amd64" --raw | audit2allow -M FTLMultiverse && sudo semodule -i FTLMultiverse.pp```
+4. Verify that the rule loaded correctly:  
+```$ sudo semanage module -l | grep FTLMultiverse```  
+You should see something along the lines of:  
+```FTLMultiverse             400       pp```  
+Then, try running FTL again. It should load with Hyperspace/Multiverse/whatever you have installed.


Nasa62:
Ideally yes, it needs to be fully tested.
And yeah, I know I don't really have good instructions on how to run jekyll in there, the basic steps I first went through are just https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/testing-your-github-pages-site-locally-with-jekyll so it should be something like installing the bundle & then running jekyll locally.

And for a website change like this, show a screenshot of the change locally so we can see what it renders like.

Nasa62:
@The-Dumb-Dino we could use translations for this for German, and Dino if anyone is still around for Korean we should ping them.

_i18n/en/install-guides/linux/index.md
```diff
@@ -35,3 +35,19 @@ To potentially fix a corrupted save:
 5. Move your original `hs_mv_prof.sav` back where it was before, replacing the one that was created by the game
 6. If this still doesn't work, delete the entire `~/.local/share/FasterThanLight/` folder
 7. ALso disable Steam Cloud Sync if you are using a version of FTL from steam, it'll frequently break the save files.
+
+#### Running FTL gives an error along the lines of "Failed to install hook"
```
Nasa62, line 39: Maybe a big `#### Fedora users see below`

Nasa62: Or actually maybe there should be a link to this section inside the install part saying
In the `linux/genericlinux/steam-install.md` and `linux/genericlinux/gog-install.md`
```md
[Fedora and SELinux users, an additional step is required](../../#selinux-steps)
```
_(I'm not 100% on the `../../` url part, might need to instead just include a template with this SELinux section so it shows up in both troubleshooting & genericlinux install guides)_

Then above your troubleshooting section here:
```md
<a name="selinux-steps" />
#### Running FTL gives an error along the lines of "Failed to install hook"
```
So that we have an easy to use named anchor rather than this giant one of the current header (and so it makes more sense during translation)


OR...
Maybe just have a section of this also mentioned there... but it should be called out at least somehow as part of the install guide, not just as troubleshooting.

kali-muon: > Maybe a big `#### Fedora users see below` ? What do you think?

Sure, that could work well. I also think that putting it in the install guide would be good; that way, people preemptively fix it while installing. Maybe we do `#### Fedora/CentOS users see below` in the install guide, then the `Running FTL gives an error along the lines of "Failed to install hook"` in the troubleshooting section.

Nasa62: Yeah lets do the callout inside the install guide, it only matters for the genericlinux gog & steam guides and not for the SteamDeck guides.

The-Dumb-Dino:
> @The-Dumb-Dino we could use translations for this for German, and Dino if anyone is still around for Korean we should ping them.

Alright I will translate when I can

ranhai613:
The homepage is renewed and SELinux exception is finally mentioned [here](https://ftl-hyperspace.github.io/FTL-Hyperspace/en/linux/troubleshooting/#the-game-fails-to-start-with-failed-to-install-hook)

PR #270, opened 2024-04-02
Exposed AddCrewMember functions for ShipManager

Choosechee:
Specifically, AddCrewMemberFromBlueprint and AddCrewMemberFromString. Although you can give crew via events, if you want something to give many different types of crew, each for a certain situation, you would have to make an event for every crew type. These functions would greatly reduce the work required for this.


ChronoVortex (approved):
Noting that setting `intruder` to `true` seems to make the new crew member replace an existing allied one, but otherwise seems to work as expected.

PR #272, opened 2024-04-02
add community korean translation

ChronoVortex:
Taken from the branch on @nedsociety's fork linked in #48.

Note that Matthew Davis has asked that any fan translations added to Hyperspace be clearly marked as such, so we'll probably need to add a note to the readmes, the page on the subset forums and the installation guide to cover all our bases.


The-Dumb-Dino:
should I maybe add this to the custom credit Screen thing?

The-Dumb-Dino:
I just realized UI elements also need to translation, I can take care of this in the future, wont have time to make it before the 1.11.0 release

PR #273, opened 2024-04-07
Removed _shortName from FunctionDefinition/Hook

bibo38:
It is unused. Also the strncpy used a bigger size, which just by luck didn't cause any problems.


Nasa62:
@bibo38 Why? Was it literally just never used? Wonder if it's a remnant from Issac.

Nasa62 (approved):
lol good catch of what could be a buffer overflow and/or segfault.

PR #276, opened 2024-04-08
Properly excluded Discord from the build if disabled

bibo38:
Previously the Discord library was only searched and linked on Windows builds. But due to the include list of DiscordIntegration.h a linux build required the discord-rpc header file to be present for a successful compilation.

Now the discord library can be enabled or disabled via a CMake option. According to the previous behaviour this option is only enabled by default on Windows builds.


Nasa62 (approved):
Looks fine, although we probably want to upgrade that Discord thing to the cross platform one in the future anyways

bibo38:
I also like having the ability to disable certain optional libraries, as it lowers the compile (and sometimes the runtime) dependencies.

PR #277, opened 2024-04-08
Changed add_custom_command to add_custom_target

bibo38:
As the custom command requires an output file, CMake seems to update the file modification time when the custom comamnd is run, therefore recompiling every dependent file. Therefore switching to a custom target, which just allows us to run a command before CMake compiles our main target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -277,19 +277,21 @@ target_sources(
 	lua/swigluarun.h
 )
 
-add_custom_command(
-	OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/lua/swigluarun.h.temp1 ${CMAKE_CURRENT_SOURCE_DIR}/lua/swigluarun.h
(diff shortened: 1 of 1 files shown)


Nasa62 (commented):
I'm not familiar with this in CMake. But if it does what it says I'm fine with it. @ChronoVortex could you test this?

ChronoVortex:
> I'm not familiar with this in CMake. But if it does what it says I'm fine with it. @ChronoVortex could you test this?

I have no idea how to test this, requested a review from you because I was hoping you'd be more capable than me.

bibo38:
FYI: My local reproduction/problem was, that after compiling everything (and keeping the build directory) calling make again (i'm using the Makefile generator of CMake, didn't test if it's the same for Ninja) compiles many files again. Therefore I tested, why they were recompiled and ended up changing `add_custom_command` to `add_custom_target`.

PR #280, opened 2024-04-12
add shortcut to access the CApp instance in lua

ChronoVortex:
Allows modders to shorten `Hyperspace.Global.GetInstance():GetCApp()` to just `Hyperspace.App`.


Nasa62:
Does this need to be added to the build/linker? I thought in the past every time we add a CPP file we need to

ChronoVortex:
This doesn't add any CPP files, I just used the existing `Misc.h` and `Misc.cpp`

PR #284, opened 2024-05-01
Exposed LocationEvent::loc; SectorDescription::type; the ChoiceReq; group and LocationEvent::Choice::requirement;

gabrielsubzero:
Exposed those two functions which should allow you to both find which sector you are currently in right now through lua and perhaps manipulate the beacon spawning system to make special events generate to the left of the map so they don't end up in places you can't reach.

Edit: Also exposed the ChoiceReq; group variables, as well as the .requirement from locationEvent, this should allow you to manipulate blue options and make lua changes to them.

### DIFF
--- a/lua/modules/hyperspace.i
+++ b/lua/modules/hyperspace.i
@@ -493,6 +493,7 @@ playerVariableType playerVariables;
 %rename("%s") SectorDescription;
 %rename("%s") SectorDescription::name;
 %rename("%s") SectorDescription::shortName;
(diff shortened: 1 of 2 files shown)


THETASOLA (approved):
All newly exposed elements are working, choiceReqs element cannot be modified for the moment but https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/297 will fix that issue.

ChronoVortex (approved):

PR #287, opened 2024-05-21
fixing ACH command and adding ACH_LOCK command

THETASOLA:
fixing ACH command and adding ACH_LOCK command.

Created a new method in the achievement tracker to remove achievements. 
`void CustomAchievementTracker::RemoveAchievement(const std::string &name)`

This method should not be exposed to lua, the ACH_LOCK command should be good enough for dev.


ChronoVortex (changes requested):
The `ACH` command works, but `ACH_LOCK` doesn't seem to. Tested with `ACH_ACC_DATABASE` in Multiverse.

ChronoVortex (approved):

ChronoVortex (approved):

PR #290, opened 2024-06-02
Added hullDamageResistChance which acts like sysDamageResistChance but for hull damage

gabrielsubzero:
Full support for this new function, the current resist icon is a light gray but changing it to white to be identical to the hull resist icon is trivial (just deleting the r,g and b color codes from the damagearea and damagebeam functions.

People who helped make this possible: doom, pepson, codyfun, chooseche and lizzard. Make sure they are mentioned in the change-log.

### DIFF
--- a/CustomShips.cpp
+++ b/CustomShips.cpp
@@ -961,7 +961,7 @@ HOOK_METHOD(ShipManager, DamageBeam, (Pointf location1, Pointf location2, Damage
 
     int room1 = ship.GetSelectedRoomId(location1.x, location1.y, true);
     int room2 = ship.GetSelectedRoomId(location2.x, location2.y, true);
-    int willDamage = room1 != room2;
+    bool willDamage = (room1 != room2);
     if (willDamage)
     {
         auto ex = RM_EX(ship.vRoomList[room2]);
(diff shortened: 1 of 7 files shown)


CustomShips.cpp
```diff
@@ -931,6 +932,70 @@ HOOK_METHOD(ShipManager, DamageSystem, (int roomId, Damage dmg) -> void)
...
+HOOK_METHOD(ShipManager, DamageBeam, (Pointf location1, Pointf location2, Damage dmg) -> void)
+{
+    LOG_HOOK("HOOK_METHOD -> ShipManager::DamageBeam -> Begin (CustomDamage.cpp)\n")
+
+    int room1 = ship.GetSelectedRoomId(location1.x, location1.y, true);
+    int room2 = ship.GetSelectedRoomId(location2.x, location2.y, true);
+    bool willDamage = (room1 != room2);
+    if (willDamage)
+    {
+        auto ex = RM_EX(ship.vRoomList[willDamage]);
```
THETASOLA, line 975: Using the variable willDamage as the room index will always return you 1, meanning that only the `hullDamageResistChance`  of Room 1 will be taken into account for a damaging beam hull resist.

You have to dynamically take into account which room is being damaged at the given tick (it always start with room1, then room2 for the rest of the beam damage tick), this condition takes care of that.
```suggestion
    if (room1 != room2)
    {
        auto ex = room2 != -1 ? RM_EX(ship.vRoomList[room2]) : RM_EX(ship.vRoomList[room1]);
```

THETASOLA (commented):
You need to update the wiki to add
`Room_Extend::hullDamageResistChance;`
and
`RoomDefinition::hullDamageResistChance;`

ChronoVortex (approved):

PR #297, opened 2024-07-09
fixing CHOICEBOX callback not being called reliably, adding :GetChoices() for LocationEvent/ChoiceBox and modifying parsefuncs.lua 

THETASOLA:
Fixing `PRE/POST_CREATE_CHOICEBOX` not being called on
 the first event and on reload

Adding `LocationEvent::GetChoices()` a method to acquire `std::vector<Choice*>` to allow lua operation
Adding `ChoiceBox::GetChoices()` a method to acquire `std::vector<ChoiceText*>` to allow lua operation

`parsefuncs.lua` had to be modified with a new feature to allow for `LocationEvent::GetChoices()`.
It now reads in the .zhl the struct `depends` for a child struct (parent::child).
image
If so, it is stored in the table/array `overridenChild`, upon generating a child struct if it is recognised to be in `overridenChild` it will not be generated, allowing for regular custom code to implement it in the wished order.


libzhlgen/lua/parsefuncs.lua
```diff
@@ -479,15 +481,15 @@ for k,fd in pairs(tfiles) do
...
             if s then
                 if func.generic_code then
                     s.generic_code = (s.generic_code or "")..func.generic_code
                 end
-                
+
                 if func.depends then
                     for _,d in ipairs(func.depends) do
                         local dname = d:cname()
+						if (string.find(dname, func:cname().."::")) then overridenChild[dname] = true end
```
Nasa62, line 492: Incorrect whitespace mess 🥇

THETASOLA: Turns out the entire parsefuncs.lua file is made of a mix of 4 space and Tab whitespace mess. I fixed them all for a 4 space indent.

ChronoVortex (approved):

Issue #327, opened 2024-08-28
drone recovery arm can work on hull repair drone


ChronoVortex:
If you jump away from a beacon while the hull repair drone is still repairing your ship, the drone recovery arm will recover the drone part. Should be fixed with an enableable setting in `hyperspace.xml`.

PR #330, opened 2024-09-03
Fixing free hull repair exploit with the repair drone and the recovery arm

THETASOLA:
Relate to issue https://github.com/FTL-Hyperspace/FTL-Hyperspace/issues/327

Sigs have only been tested on win32.

Also the way the linux and windows version handle the recovery arm is different so linux might not work on testing.


ChronoVortex (approved):
Verified on Linux

PR #333, opened 2024-09-05
fixes for overflowing enemy HP bar

THETASOLA:
Two new addition to the handling of enemy Hull Bar visual indicator.

Previously enemy with more than 22 hp would break the UI by having their hp bar render over the ship name and status until overflowing out of the screen.

Now with the default option the HP bar will visually stops at that 22 limit (can be set in the xml) and scale it down according to the  enemy max HP. (so a starting 100hp enemy at 50hp will visually show the bar as it is it was at 11hp in the interface, if the limit is untouched)

There is also another option that "takes the overflow" to make it a layered HP bar system like many recent game
image
The colours of each layer can be set in the xml.

The sig for CachedImage::SetPartial is untested on linux.

### DIFF
--- a/FTLGameELF32.cpp
+++ b/FTLGameELF32.cpp
@@ -2991,6 +2991,20 @@ void CachedImage::SetScale(float wScale, float hScale)
 namespace _func215
 {
     static void *func = 0;
+	static short argdata[] = {0x1ff, 0x1ff, 0x1ff, 0x1ff, 0x1ff};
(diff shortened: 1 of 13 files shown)


ChronoVortex (changes requested):
Validated on Linux, but `type="2"` not working as intended

ChronoVortex (approved):
Turns out I was completely wrong, both methods work as intended

PR #334, opened 2024-09-06
fix for arty vanishing from gibs

ChronoVortex:
Also fixed some of `ExplosionAnimation`'s members not having parity with the vanilla struct


Nasa62 (approved):

ChronoVortex:
Verified on Linux

THETASOLA (approved):

PR #346, opened 2024-09-14
CrewMember::OnRender callback

THETASOLA:
CrewMember::OnRender callback

To be tested for 1.15.
TBD if it's also relevant and not too resource intensive to handle.

### DIFF
--- a/Misc.cpp
+++ b/Misc.cpp
@@ -1430,8 +1430,20 @@ HOOK_METHOD(Ship, OnRenderJump, (float progress) -> void)
     lua_pop(context->GetLua(), 2);
 }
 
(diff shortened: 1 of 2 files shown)


The-Dumb-Dino:
Change the 
auto context = Global::GetInstance()->getLuaContext();
to
LuaScriptInit *context = Global::GetInstance()->getLuaContext();

We don't like auto

THETASOLA:
Then to be consistent we'll have to update every callback to that form.
This is a change that must be done on a separate PR.

The-Dumb-Dino:
>  This is a change that must be done on a separate PR.
Good call, I wanted to make a "remove most auto's PR at some point either way.

Issue #317, opened 2024-08-07
Cannot input Japanese letters into vanilla text box


ranhai613:
This happens when renaming ship name in hangar and crew name. The default Japanese name is written in the text box, but I can only add latin letters or deleting existing Japanese letters.

THETASOLA:
Handled by #357, thanks ranhai

PR #357, opened 2024-09-19
Allow rename input Japanese letters (fix issue 317)

ranhai613:
Fix issue #317 

Pepson gave me the hook method to fix the problem. This PR makes the crew and ship rename input accept Japanese letters (or more letters).


THETASOLA:
This would also allows special character in ship/crew name, there might be a need to make it an option.

THETASOLA (approved):

PR #359, opened 2024-09-25
Equipment slots visualization for not installed weapon or drone system

ranhai613:
{6D919E65-538E-4B47-9B28-4FB6EFE47DAA}

Add transparent equipment slots under the text "SYSTEM NOT INSTALLED" in hangar and equipment menu. Players can see how many slots a weapon or drone system has before installing it.

I only tested on win32, so not sure sigs for arm64 and x84 on Linux work.


ranhai613:
I integrated this features into #416 with not hacky way.

PR #361, opened 2024-09-27
Custom choice text color

ranhai613:
Have not tested on Linux.

If evrything is okay, I will open another pr for lua implemention.


CustomColors.cpp
```diff
@@ -318,3 +322,170 @@ HOOK_METHOD_PRIORITY(FTLButton, OnRender, 5000, () -> void)
...
+
+void ParseChoiceColorNode(rapidxml::xml_node<char>* node)
+{
+    if (node->first_attribute("name") == nullptr) return;
+
+    std::string name = std::string(node->first_attribute("name")->value());
+    if (ChoiceColorMap[name] != nullptr) return;
+
+    ChoiceColor* choiceColor = new ChoiceColor;
+    if(node->first_node("normal") != nullptr)
```
ChronoVortex, line 343: There should be a space between the `if` and the opening parenthesis. You also shouldn't need to check `!= nullptr`, `node->first_node("normal")` will evaluate as `true` on its own if it's not null. All instances of these two issues should be fixed.

CustomColors.cpp
```diff
@@ -318,3 +322,170 @@ HOOK_METHOD_PRIORITY(FTLButton, OnRender, 5000, () -> void)
...
+HOOK_STATIC(freetype, easy_printAutoNewlines, (int fontSize, float x, float y, int line_length, const std::string &text) -> Pointf)
+{
+    LOG_HOOK("HOOK_STATIC -> freetype::easy_printAutoNewlines -> Begin (CustomColors.cpp)\n")
+    if (!CustomOptionsManager::GetInstance()->enableCustomChoiceColors.currentValue) return super(fontSize, x, y, line_length, text);
+
+    std::regex re("^(.*)\\[\\[#C\\:(.*?)\\]\\]$");
+    std::smatch match;
+    std::string new_text = text;
+    GL_Color currentColor = CSurface::GL_GetColor();
+    if (std::regex_match(new_text, match, re))
```
ChronoVortex, line 453: I'm not a huge fan of all these regex checks being made every frame in all of these text printing functions, it seems pretty expensive. Would it be possible to just rewrite the function that renders event choices instead?

ChronoVortex (changes requested):

CustomColors.cpp
ChronoVortex: New check makes it much better, but could be improved in the future.

ChronoVortex (changes requested):
Signatures work on Linux, but the regex does not.

ChronoVortex (approved):
Seems to be fixed on Linux

PR #363, opened 2024-10-03
Enemy Charge Box Position Fix

ranhai613:
Fix the issue from [Enemy ships where rightmost weapon charge bar is not visible due to being outside the screen](https://discord.com/channels/604415384979898464/1244130530014920716)

Have not tested linux sigs.

Prevents enemy charge bar (sensor lvl 3+ ability) from framing off the right, top and bottom side of the screen.

Before:
image

After:
image

I tested following MV ships that have the problem:
- MU_REBEL_MARAUDER
- MU_DYNASTY_BOMBER
- MU_FREEMANTIS_DROPSHIP
- MU_REBEL_BATTLESHIP
- MU_AUTO_DREAD
- MU_SUZERAIN_JAILER
- MU_LEECH_BOMBER
- MU_ANGEL_FIGHTER

~~There is another problem due to the fix: the moved charge bar overlaps with the next one.~~
image

~~Although we need to figure out a way to get the absolute position of the box in order to fix it, the current fixed box still looks fine in most cases.~~
Fixed.


ranhai613:
For now I will leave this a draft. I'm going to find a complete fix.

ranhai613:
Pursuiding the absolute positions, I found it depends on the ship position. So I need to get the absolute position of the ship. Maybe I could find the way if I continue to investigate, though if anyone know it, please tell me so that I can save my time.

THETASOLA:
Confirmed sigs on linux.

THETASOLA (approved):

Issue #350, opened 2024-09-16
Custom achievements use the wrong sound


JerseeBeep:
Custom achievements will play the "Ship unlocked" sound rather than the base achievement unlock sound.

ChronoVortex:
The sound that achievements use should probably be configurable in the xml.

PR #380, opened 2024-10-18
Custom sound for achievements

THETASOLA:
Required adding `AchievementTracker::OnLoop()` sig.
Untested on linux.

Adds the `<sound>sound_id</sound>` tag to the `<achievements>` xml, allowing for custom sounds.
Also somewhat fixes the issue of `<header>` replacing the achievements sound by the ship unlock sound when uneeded. #350


ChronoVortex (approved):
Validated on Linux

PR #381, opened 2024-10-20
Custom in-text color

ranhai613:
Modders can set In-text color, on particall characters with colors specified as RGB(A).

Within a text in \<text> tag, a part enclosed by start flag [color:RRGGBB] and end flag [/color] is painted with the color specified in start flag as hex rgb value (0-f).

To enable this feature, you need to **_sacrifice_** one letter, meaning you cannot use the letter so far and in the future. Its like how ~ or | are replaced with fuel or drones symbols in MV (in MV you cannot use the letter as it is).

How to sacrifice:
You have to modify font and make one letter 1pixcel-wide and invisible. You will use [this tool](https://github.com/D-side/ftl_font) to modify .font file. As a basic usage, please refer to the readme of the site.

1. Dismantle the target font file.
2. Choose one letter to sacrifice. It cannot be used as a letter once enabling this feature, so you  should pick out a letter that is not used as far and would not in the future. Or you can add a new letter to the font I guess?
3. Overwrite the font. Follow the readme. Make 1 pixel square image. The color must be black. When setting propaties of the letter in index.json, set the value of two items of the letter `before` and `after` to 0.
4. Reassemble and put the product into `fonts/` of your mod.

Inner workings:
The hook method does easy_print multiple times with different colors. Each color has corresponding text where the specified part by the flags are left, but other parts are filled with 1 pixel-wide invisible characters so that painting would not coflict. To adjust the text position, I use easy_measureWidth to measure pixels and fill up with invisible letters.

Here is the modded font I use. I replaced `§` with a 1 pixel-wide invisible letter.
https://drive.google.com/file/d/1SqrGBm7Ke6Q2Ja9SmQP9O_wDsQO7aLIG/view?usp=sharing

### DIFF
--- a/CustomColors.cpp
+++ b/CustomColors.cpp
@@ -1,6 +1,8 @@
 #include "CustomColors.h"
 #include "Resources.h"
 #include "PALMemoryProtection.h"
+#include <regex>
+#include <codecvt>
 
 GL_Color g_defaultTextButtonColors[4] =
 {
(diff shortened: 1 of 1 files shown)


ranhai613:
I have come up with another way to implement in-text color. For now I close this pr and will open another pr when it is ready.

PR #386, opened 2024-10-31
Exposed LockdownShard

gabrielsubzero:
As the title says, exposed the LockdownShard function and all the subfunctions from it which should allow for custom lua lockdowns.

### DIFF
--- a/FTLGameELF32.h
+++ b/FTLGameELF32.h
@@ -6717,6 +6717,17 @@ struct Ship : ShipObject
 		return graph->GetRoomCenter(room);
 	}
 
+	std::vector<LockdownShard*> GetShards()
+    {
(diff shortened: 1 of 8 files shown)


ranhai613 (changes requested):
It seems all field are virtually immutable by a weird behavior in base code. You can set the value but it is immediately set back, so any change in lua makes no effect. We need to inspect it more eventually, for now you should note on wiki that you cannot change the value. Also `Update` method makes no sense for the same reason. The method moves shard forward during flying and ages it. If you are not interested in it, you can just remove it from this pr.

ranhai613 (approved):

THETASOLA (approved):

Issue #372, opened 2024-10-16
crew clone animations don't render with color layers


ChronoVortex:
From what I can tell, `CloneSystem::OnRenderFloor` is what handles this. It only renders `currentCloneAnimation`, which presumably loads the clone animation for whichever crew is being cloned. This only has the base layer, so this function would need to be rewritten to include rendering the color layers.

The-Dumb-Dino:
if its not too much I'd do that

PR #392, opened 2024-11-10
Custom Clonebay

THETASOLA:
Adds 6 new xml parameters to the clonebay system, that dictates

- Cloning speed
- Clone death speed (when unpowered)
- Skill loss on clone
- HP gained per jump
- % of HP gained per jump
- % of HP when cloned

For this I had to add a few new sigs (only for win32 currently)

- `CloneSystem::CloneReady`
- `CloneSystem::GetJumpHealth`
- `CloneSystem::GetDeathProgress`
- `CloneSystem::GetProgress`
- `CloneSystem::GetCloneTime`

There's also a new `CREW_CLONE` callback

A quick quirk of the change is that clonebay level 4 and above cannot have a cloning time above 0, it's just a limitation of the code of ftl that can only be resolved with a rewrite of the loop.

There is also the xml groundwork for the number of crew that can be revived at the same time, but to achieve that someone will probably have to rewrite the render of the clone system.


ChronoVortex:
Linux signatures still need to be added. If we rewrite the render event for the clonebay we should fix #372 while we're at it.

ChronoVortex (approved):
We've done all we can, time to hand it to the playtesters

PR #394, opened 2024-11-17
Add overloads for hs_log_file

ranhai613:
I often use hs_log_file as printing values, though it can only handle *char. So this pr add basic overloads for hs_log_file so that it can help me a lot for HS development.

Add overloads for following types:
- std::string
- bool
- int
- unsigned int
- long
- unsigned long
- long long
- unsigned long long
- float
- double
- long double

### DIFF
--- a/Global.cpp
+++ b/Global.cpp
@@ -46,6 +46,32 @@ void hs_log_file(const char *str...)
     va_end(va);
 }
 
+void hs_log_file(const std::string &str)
+{
+    FILE* log = G_->logFile;
+    if (log != nullptr)
+    {
+        fputs(str.c_str(), log);
+        fflush(G_->logFile);
+    }
+}
+
+void hs_log_file(bool val)
+{
+    if (val) hs_log_file("true");
+    else hs_log_file("false");
+}
+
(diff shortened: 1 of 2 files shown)


ChronoVortex:
`hs_log_file` is based on `printf`, so it can certainly already handle a lot of the values you've overloaded for if you format the string correctly. I'd recommend having a look at the `printf` docs for C++.

Regardless, the overload for `std::string` is useful, so we can probably at least keep that.

ranhai613:
> `hs_log_file` is based on `printf`, so it can certainly already handle a lot of the values you've overloaded for if you format the string correctly. I'd recommend having a look at the `printf` docs for C++.
> 
> Regardless, the overload for `std::string` is useful, so we can probably at least keep that.

Yeah at that point I didnt realize I can use formatted string lol. Actually now Im taking advantage of it and feel very comfortable. This pr would not benefit me much, so I will leave it to you if you merge this pr. Feel free to discard if you wont.

THETASOLA (approved):

THETASOLA:
I'm keeping the overload for std::string because that is good QoL. The rest is can'.

PR #397, opened 2024-11-22
Fix save file breaks when saving whilst a weapon still has queued projectiles

### LINKED ISSUE #378: Save file breaks when saving whilst a weapon still has queued projectiles (We now finally know whats the infamous save breaker of MV lol)
Through some testing of my own code I found out that saving the game whilst any weapon is in its shooting phase (means that it still has one or two out of three total projectiles waiting to be shot) then the game will crash upon loading the continue save file.

I am unsure what exactly causes that but I was able to track down where is issue originates by commenting out some functions and filtering when the game would load the continue save. 
These two saving hooks are the criminals here:
```c++
HOOK_STATIC(ProjectileFactory, LoadProjectile, (int fh) -> Projectile*)
{
    LOG_HOOK("HOOK_STATIC -> ProjectileFactory::LoadProjectile -> Begin (CustomDamage.cpp)\n")
    Projectile *ret = super(fh);

    if (!ret)
    {
        hs_log_file("ProjectileFactory::LoadProjectile failed!");
        return ret;
    }

    Projectile_Extend *ex = PR_EX(ret);

    ex->customDamage.sourceShipId = ret->ownerId;

    ex->name = FileHelper::readString(fh);

    ex->customDamage.def = CustomDamageDefinition::customDamageDefs.at(FileHelper::readInteger(fh));
    ex->customDamage.accuracyMod = FileHelper::readInteger(fh);
    ex->customDamage.droneAccuracyMod = FileHelper::readInteger(fh);

    int n = FileHelper::readInteger(fh);
    for (int i=0; i<n; ++i)
    {
        ex->missedDrones.push_back(FileHelper::readInteger(fh));
    }

    return ret;
}

HOOK_STATIC(ProjectileFactory, SaveProjectile, (Projectile *p, int fh) -> void)
{
    LOG_HOOK("HOOK_STATIC -> ProjectileFactory::SaveProjectile -> Begin (CustomDamage.cpp)\n")
    super(p, fh);

    Projectile_Extend *ex = PR_EX(p);

    FileHelper::writeString(fh, ex->name);

    FileHelper::writeInt(fh, ex->customDamage.def->idx);
    FileHelper::writeInt(fh, ex->customDamage.accuracyMod);
    FileHelper::writeInt(fh, ex->customDamage.droneAccuracyMod);

    FileHelper::writeInt(fh, ex->missedDrones.size());
    for (int i : ex->missedDrones)
    {
        FileHelper::writeInt(fh, i);
    }
}
```
You can confirm this yourself by saving with these two being commented out if you don't trust me.

Thanks to whoever takes care of this, I really need this to be fixed, otherwise I can't continue with my saving rewrite. The MV community will also thank you a lot for having a more stable experience.

ranhai613:
fix #378 
Have not tested on linux

This bug is due to Vanilla FTL devs' crime. Projectiles in space are saved within SpaceManager::SaveSpace and loaded within SpaceManager::LoadSpace, and queued projectiles are saved within ProjectileFactory::SaveState and loaded within ProjectileFactory::LoadState. Both loading functions call ProjectileFactory::LoadProjectile which is hooked by HS to load projectile_extended stuff. However, although SpaceManager::SaveSpace calls ProjectileFactory::SaveProjectile which is also hooked by HS for projectile_extended, ProjectileFactory::SaveState does NOT call it even though a part of it does the exactly same thing as ProjectileFactory::SaveState (there is a possibility that the compiler did inlining). Due to not calling it, the info of extended stuff of queued projectiles are not saved, but HS trys to load them. This is why the crash occurs.

This pr rewrites ProjectileFactory::SaveState and comments out the crime part, instead calls ProjectileFactory::SaveProjectile so that the saving and loading process correctly proceed.

Thanks to Dino, finding out where the crash comes from

### DIFF
--- a/CustomDamage.cpp
+++ b/CustomDamage.cpp
@@ -452,6 +452,55 @@ HOOK_STATIC(ProjectileFactory, SaveProjectile, (Projectile *p, int fh) -> void)
     }
 }
 
+HOOK_METHOD_PRIORITY(ProjectileFactory, SaveState, 9999, (int fd) -> void)
+{
+    LOG_HOOK("HOOK_METHOD -> ProjectileFactory::SaveState -> Begin (CustomDamage.cpp)\n")
+
+    // Reverse engineered Vanilla code
+    FileHelper::writeFloat(fd, cooldown.first);
+    FileHelper::writeFloat(fd, cooldown.second);
+    FileHelper::writeFloat(fd, subCooldown.first);
+    FileHelper::writeFloat(fd, subCooldown.second);
+    FileHelper::writeInt(fd, boostLevel);
+    FileHelper::writeInt(fd, chargeLevel);
+    FileHelper::writeInt(fd, targets.size());
+    for (auto pointf : targets)
+    {
+        FileHelper::writeInt(fd, (int)(pointf.x));
+        FileHelper::writeInt(fd, (int)(pointf.y));
+    }
+    FileHelper::writeInt(fd, lastTargets.size());
+    for (auto pointf : lastTargets)
+    {
+        FileHelper::writeInt(fd, (int)(pointf.x));
+        FileHelper::writeInt(fd, (int)(pointf.y));
+    }
+    FileHelper::writeInt(fd, (unsigned int)autoFiring);
+    FileHelper::writeInt(fd, (unsigned int)fireWhenReady);
+    FileHelper::writeInt(fd, targetId);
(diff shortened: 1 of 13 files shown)


THETASOLA (approved):

THETASOLA:
Waiting on @ChronoVortex confirmation for the linux sigs.

Issue #398, opened 2024-11-22
reactor upgrade bug


localresources:
Currently on a ship with a reduced reactor cap (an example of such being zoltan c, the reactor upgrade UI initially shows the reactor cap as the standard 25, but after a certain point (from my experience trying to upgrade reactor from 16 to 17) the reactor UI suddenly switches to show the proper reactor caps. Ideally it would show the correct reactor cap regardless of how much it has been upgraded.

PR #399, opened 2024-11-23
Fix upgrade rendering for reduced reactor cap

ranhai613:
Fix issue #398 
change the original conditon
`else if((tempLevel > (maxLevel - 4)) && (currentBar > displayMaxLevel)) continue;`
to
`else if (currentBar > displayMaxLevel) continue;`

Not sure what `(tempLevel > (maxLevel - 4))` is intended, only laszlogasd who wrote this would know.

And make columns that have no upgrade rectangles due to low max power display `--` at the price display instead of increased value form previous price.

~~Also, this hook seems never call super inside and intend to rewrite vanilla function. So I replace HOOK_METHOD with HOOK_METHOD_PRIORITY with priority 9999. This is what rewriting should be.~~
Reverted to HOOK_METHOD

### DIFF
--- a/CustomReactor.cpp
+++ b/CustomReactor.cpp
@@ -72,7 +72,7 @@ HOOK_METHOD(ReactorButton, OnRightClick, ()->void)
     }
 }
 
-HOOK_METHOD(ReactorButton, OnRender, ()->void)
+HOOK_METHOD_PRIORITY(ReactorButton, OnRender, 9999, ()->void)
 {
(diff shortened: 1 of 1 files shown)


THETASOLA:
> Also, this hook seems never call super inside and intend to rewrite vanilla function. So I replace HOOK_METHOD with HOOK_METHOD_PRIORITY with priority 9999. This is what rewriting should be.

This is not necessary, `HOOK_METHOD_PRIORITY` is only used when you have multiple `HOOK_METHOD` and you want to be sure that one is ran before the other.
Regular `HOOK_METHOD` will always run before the vanilla function.

THETASOLA (approved):

PR #402, opened 2024-11-28
fix move command on partitioned ships

### LINKED ISSUE #339: use partition system to fix move command on partitioned ships
When ordering a crew member to move to a room it can't reach, it will force a crew member on the same tile away, but won't go there itself. It should instead go to the backup room defined in the partition system from #332.

### LINKED ISSUE #341: clean up remaining return to stations edge-cases
Despite the fix in #332, there are still some situations which will cause crew to phase out because they are unable to reach their station. Example from playtesting:

> Just got my slug pilot and my rock pilot to occupy the same space with the aid of the chaotic surgeon drone. If crew A is stationed at piloting, crew B is currently manning piloting, and the drone is in crew B's square in the piloting room, then crew A and crew B will proceed to get cozy at the wheel together when I return them to their stations.

https://github.com/user-attachments/assets/0a51e4ec-ec43-44af-9fe8-cd6d15163840

This could potentially be fixed with a hook for `CrewMember::GetSavedPosition` which returns an empty slot if the crew member is unable to reach their station. This should only apply when called from inside `ShipManager::RestoreCrewPositions` so as not to affect its use in `ShipManager::ExportShip`. Example of returning an empty slot:

```c++
Slot slot = Slot();
slot.roomId = -1;
slot.slotId = -1;
slot.worldLocation = {-1, -1};
return slot;
```

THETASOLA:
All is in the title.

fix #339

### DIFF
--- a/CustomCrew.cpp
+++ b/CustomCrew.cpp
@@ -6781,7 +6781,7 @@ HOOK_METHOD(CrewMember, RestorePosition, () -> bool)
         CrewMember_Extend *ex = CM_EX(this);
         ShipGraph* graph = ShipGraph::GetShipInfo(currentShipId);
         Room* stationRoom = graph->rooms[station.roomId];
-        if (stationRoom->Full(false)) return false; // Handle the the case where a stationless crew/drone is filling the room of the saved station
+        if (stationRoom->Full(false)) return false; // Handle the case where a stationless crew/drone is filling the room of the saved station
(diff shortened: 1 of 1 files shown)


ChronoVortex (approved):
Crew still move out of the way if you order someone to a tile they can't reach, but otherwise seems fine. Don't mind making that small concession.

Issue #323, opened 2024-08-19
warning light for beaches and fires can render over doors


ChronoVortex:
It always renders in the middle of the bottom side of a room:

![tmp](https://github.com/user-attachments/assets/b68afe0d-914f-4e86-bc81-81dab83592cb)

The rendering of this sprite is handled in `Room::OnRenderWalls`, but the position can be modified by destroying `lightPrimitive` and `lightGlowPrimitive` in a hook for the `Room` constructor, then recreating them with the correct position and rotation.

PR #403, opened 2024-11-28
Fix warning lights position

ranhai613:
Fix issue #323 

This pr adds an algorithm to relocate warning lights so that they no longer stick to doors or leave from walls.

Example 1 Before
image
Example 1 After
image

Example 2 Before
image
Example 2 After
image


ChronoVortex:
Fantastic, been itching for this one for a while, thanks ranhai! Looking forward to reviewing

ChronoVortex (approved):

Issue #324, opened 2024-08-22
clone sound can play on enemy ship after it's been destroyed


ChronoVortex:
Still seems to be an issue despite the fix in 0c3a7b11dcf58f2ecb4163b543d5a1f0de1f4aa5

PR #404, opened 2024-12-02
Fix clone sound playing for destroyed ships (again)

ranhai613:
Fix issue #324 
MAYBE fixed. I could not test because not sure how to replicate the bug ¯\\_(ツ)\_/¯
I would appreciate you if you tell me how to replicate.


ChronoVortex:
Unfortunately I'm not entirely certain how to replicate the issue. It came up in one of my runs on the latest release and I assumed it'd be easy, but if you weren't able to figure it out I suppose not. I think I'll take this off the milestone for now and hold it in reserve until we know for certain what the cause is.

Issue #395, opened 2024-11-18
Show Internal Event Name on Error


neopryne:
The current error message ```Failed to call the callback for InternalEvent X``` could easily be more helpful if it also printed the name of internal event X.  This could be added with a lookup table from numbers to strings.

PR #419, opened 2025-01-12
Display enum name when getting lua error

ranhai613:
for #395

This pr takes new enum item of InternalEvents from custom clone bay pr into account beforehand. So auto build error due to assert is intentional.


ranhai613:
#421 is the superior alternative implementation

THETASOLA:
Considering the existence of #421 I am closing this one.

PR #421, opened 2025-01-14
Improved lua errors

sillysandvich:
-Alternative implementation of pr/419
-Handles issue #395
-Automatically maps InternalEvents/RenderEvents enums to readable strings -Adds checking at function registration time for proper number of arguments


Misc.cpp
```diff
@@ -801,16 +801,34 @@ int LuaLibScript::l_on_internal_event(lua_State* lua)
...
-    printf("Registered Lua Function: that accepts '%u' arguments and is variable arguments: %s\n", ar.nparams, ar.isvararg ? "TRUE" : "FALSE");
-    */
-
+    
     InternalEvents::Identifiers id = static_cast<InternalEvents::Identifiers>(callbackHookId);
+    InternalEvents::EventInfo info = InternalEvents::GetEventInfo(id);
+    const char* eventName = InternalEvents::GetName(id);
+    if (ar.nparams != info.argCount)
+    {
+        hs_log_file("Error: Callback function for InternalEvent %s has the wrong number of arguments! Expected %u, got %u\n", eventName, info.argCount, ar.nparams);
```
ranhai613, line 817: It would be better to show error message box as well as log the error by hl_log_file. So the error attracts more attention.

like:
`luaL_error(lua, "Error: Callback function for InternalEvent %s has the wrong number of arguments! Expected %d, got %d", eventName, info.argCount, ar.nparams);`

Misc.cpp
```diff
@@ -801,16 +801,34 @@ int LuaLibScript::l_on_internal_event(lua_State* lua)
...
+        hs_log_file("Expected function of the form: %s\n", info.functionSignatureDescription);
+        return 0;
+    }
+    else if (ar.isvararg != info.isVariableArgs)
+    {
+        hs_log_file("Error: Provided function for InternalEvent %s %s!\n", eventName, info.isVariableArgs ? "Should be a variable argument function" : "Should not be a variable argument function");
```
ranhai613, line 823: so here too

lua/InternalEvents.h
```diff
@@ -213,5 +188,4 @@ struct BeamHit
         NEW_ROOM   // treat beam hit as new room from last frame
     };
 };
-
-#endif // INTERNALEVENTS_H
+#endif // INTERNALEVENTS_H
```
ranhai613, line 191: It needs a linebreak here too.
```suggestion
#endif // INTERNALEVENTS_H

```

lua/EventEnums.h
```diff
@@ -0,0 +1,26 @@
...
+    };
+#else //#ifdef SWIG
+    %define NAMED_ENUM(enumName, ...)
+        enum enumName {
+            __VA_ARGS__
+        };
+    %enddef //NAMED_ENUM
+#endif //#ifdef SWIG
+
+#endif //EVENTENUMS_H
```
ranhai613, line 26: Because of HS building process quirkiness, a linebreak is somewhat mandatory to avoid breaking some things (hence why they are in every file)
```suggestion
#endif //EVENTENUMS_H

```

ranhai613 (commented):
Thanks for opening PR for enum name things! Your implemention is much better than mine. I commented some suggestions. Im not sure when we can merge this PR because we are struggling with other PR but Im looking forward to merging this!

Misc.cpp
sillysandvich: Good idea, I added this.

THETASOLA (approved):

THETASOLA:
Seems all good on my side, I will be merging this.
We will have to watch out for all current PRs that add Internal Events, there should be a merge conflict with those but better to be safe. @ranhai613 @ChronoVortex

PR #424, opened 2025-01-16
Fix ELF segment size calculation

afishhh:
Doesn't crash if a symbol can't be found.


SigScan.cpp
```diff
@@ -258,13 +258,9 @@ static int callback(struct dl_phdr_info *info, size_t size, void *data) {
 
             programBaseAddress = (uintptr_t) (info->dlpi_addr + t_phdr.p_vaddr);
             
-            size_t seg_size = 0;
             // get segment size
-            while(seg_size < t_phdr.p_filesz) {
```
Nasa62, line 263: Okay @THETASOLA so I think this is because in the ELF spec there's multiple segments, so I was trying to go through every segment I was looking at in the file to find the total-end scanning range.

However; what it probably needs to be overall (if we're matching things in multiple segments, which we're probably not but when I was implementing this I was learning about ELF) is several different scanner ranges of starts & stops if we wanted to match across the different executable sections but that's a much bigger change to ZHL

Nasa62 (approved):
This looks like it'll be fine if we're only scanning a single-segment anyways.

For Linux this is probably true, not 100% sure about mac yet.

@The-Dumb-Dino this will be good to bring into/rebase the Mac branch if this causes us to be unable to match any necessary signatures there or if this helps us to find signatures we can't match correctly.

SigScan.cpp
afishhh: > I was looking at in the file to find the total-end scanning range

The previous code would also assign to segmentLength just with a wrong value so it wouldn't do that anyway and I think the exact same issue would occur.

afishhh:
> For Linux this is probably true, not 100% sure about mac yet.
> 
> @The-Dumb-Dino this will be good to bring into/rebase the Mac branch if this causes us to be unable to match any necessary signatures there or if this helps us to find signatures we can't match correctly.

Is this code relevant for MacOS? I thought it wouldn't provide these APIs since it uses Mach-O and not ELF, and according to stackoverflow it provides a different Mach-O specific API for this instead.

THETASOLA:
Are we waiting on something or can we merge this?

afishhh:
> Are we waiting on something or can we merge this?

I don't think there is anything blocking this, and as far as I can tell it works fine.

PR #432, opened 2025-01-20
Replacement implementation for Get_Drone_Subclass

sillysandvich:
## Overview

Automatically downcasts `SpaceDrone*` instances to their final type when passing them to Lua.

This is similar to what `Get_Drone_Subclass` did, except this approach applies more generally to `SpaceDrone*` as a type and not just the result of a `vector<SpaceDrone*>` indexing operation.

**NOTE:** This does not apply to objects passed via `SWIG_NewPointerObj` (such as within [InternalEvents::CONSTRUCT_SPACEDRONE](https://github.com/FTL-Hyperspace/FTL-Hyperspace/blob/34cd584c95270dd0d064691e5b0d04906b9f60f2/CustomDrones.cpp#L1227)), so `Get_Drone_Subclass` usage for these cases will no longer work. However, downcasting here would be a bug anyway, so this is unlikely to be an issue.


ChronoVortex:
It would be nice to have this for `Projectile` as well so that `BeamWeapon` is accessible.

sillysandvich:
Note that before merging, a regular `out` typemap would also work, the `DYNAMIC` typemap is best used for inheritence hierarchies because the conversions can chain for complex casts.

Issue #326, opened 2024-08-26
Add dummy systems or features to create them from artillery


arcburnergit:
The functionality for custom systems can be created quite easily inside of lua, however what is hard to replicate is the actual system itself.
Ideally we'd be able to define fake systems and give them a name and decide the offset between it and the next system on the bar, as well as a custom icon ect.
The other option is adding the features that would enable the same thing but created from artillery. we'd need some way to change the artillery icon, the option to hide the artillery charge time and to change the offset between artillery and the other systems on the bar.

PR #433, opened 2025-01-23
Custom Systems Implementation.

sillysandvich:
This PR addresses issue #326 and implements a framework for custom systems.
## Changes
- Modders can now add `<customSystem>` tags within `<customSystems>` in order to register valid system IDs with the game for dummy systems.
- Partially exposed SystemBox class to lua.
- Implemented and exposed SystemBox_Extend class.
- Added several `InternalEvent`s:
    -  `SYSTEM_BOX_MOUSE_MOVE`
    - `SYSTEM_BOX_MOUSE_CLICK`
    -  `SYSTEM_BOX_KEY_DOWN`
    -  `GET_LEVEL_DESCRIPTION`
- Added RenderEvent: 
    - `SYSTEM_BOX`
 - Removed some code duplication in function overloads.
 
 ## ~~Required additions~~
 - ~~SystemBox::destructor signature needs to be added for Linux platforms.~~
 - ~~FTLGameStripped headers for linux need to add appropriate padding for SystemBox struct so that SystemBox_Extend can work.~~
 
 ## Subjects to consider
 - `InternalEvents::SYSTEM_BOX_MOUSE MOVE` and `RenderEvents::SYSTEM_BOX` currently adjust the reference frames of rendering matrices and passed arguments to line up with the appropriate SystemBox. This is not done elsewhere.
 - `InternalEvents::GET_LEVEL_DESCRIPTION` currently uses callbacks to avoid additional complexity, but a map from custom system ids to individual LuaFunctionRefs may be more appropriate and efficient.
 ## Future improvements
 - Possibly allowing the user to control the ordering of systems within the UI rather than making it depend on order of declaration?
 - Implement mutually-exclusive systems (Medbay/Clonebay).
 - Rework system rendering to use `xOffset` variable for all systems and not just custom ones (To accommodate UI elements on the left side of a system).
 - Add callback for `SystemBox::GenerateTooltip` or implement some method to provide manning bonus descriptions for custom systems if necessary.
 - Rework subsystem UI and rendering to accommodate user subsystems.  
 ## Todo
 - ~~Integrate with `UpgradeBox` so the upgrade menu functions properly~~
 - ~~Distinguish between ship selection and game UI system rendering (no buttons on ship selection). This could probably be done by checking `ShipBuilder::bOpen` but it may be neater to find a way to do it within callbacks by either passing an argument or only calling during gameplay proper.~~
 Done.


The-Dumb-Dino:
> Here is a simple mod with an example of a new custom system. 
> [systems.zip](https://github.com/user-attachments/files/18518016/systems.zip)
> 

When will you finally join us on discord?

The-Dumb-Dino:
## Another possible improvement:
Make custom systems removeable mid run as it's a feature we plan to add sooner or later to vanilla systems.

ChronoVortex:
> Implement mutually-exclusive systems (Medbay/Clonebay).

It would be good to make it possible for a custom system to be mutually exclusive with a vanilla system too. Say, for example, you could purchase a system like the husk racks from Void War, but it would replace drone control.

sillysandvich:
Updated version for testing (Fixed bugs on the lua side). Enemy usage isn't handled but that is well within the capabilities of the lua interface for simple systems like this.
[systems.zip](https://github.com/user-attachments/files/18604520/systems.zip)

CustomSystems.cpp
```diff
@@ -33,7 +34,44 @@ void ParseSystemsNode(rapidxml::xml_node<char>* node)
...
+{
+    auto it = systemIds.find(systemName);
+    if (it != systemIds.end()) return it->second;
+    return SYS_INVALID;
+}
+std::string CustomUserSystems::SystemIdToName(int systemId)
+{   
+    int idx = systemId - SYS_CUSTOM_FIRST;
+    if (idx >=0 && idx < systemNames.size()) return systemNames[idx];
+    return "";
```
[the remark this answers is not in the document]
sillysandvich: I didn't catch that, that's a good idea.

ranhai613:
Do you think it is possible to add save/load mechanics for custom systems? I played your example mod and found the system state (provided reactor power, locked amount) isnt saved.

sillysandvich:
> Do you think it is possible to add save/load mechanics for custom systems? I played your example mod and found the system state (provided reactor power, locked amount) isnt saved.

I definitely think so, probably by iterating over the `vSystemList` and then calling `ShipSystem::SaveState/LoadState` for all systems with an ID that indicates that they are custom. I'll have to look into how native saving works in more detail to see if there's something more complicated that needs to be done for this.

ranhai613:
Good to know. Seeing the damages the custom system has taken arent saved, player can easily abuse this save/load glitch. We need to address saving/loading now.

ranhai613:
Seems you made it and now saving/loading is working. Great job!!

sillysandvich:
I think I have an approach to this, would you mind uploading the mod that you're using to test so I can use it as well?

ranhai613:
> I think I have an approach to this, would you mind uploading the mod that you're using to test so I can use it as well?

[s.zip](https://github.com/user-attachments/files/18689503/s.zip)

I just messed up your example mod but here. Hopefully it can help you

sillysandvich:
image
If you look closely the glows are a little bit odd, but this is definitely an improvement.
I think making offsets work for all systems would require more work if you wanted to add UI elements to the LEFT side of a custom system, but for now this should fix any glaring issues.

sillysandvich:
image
Good catch.

ranhai613:
Thanks for quick fix!

image
Do you know how to disable this portrait icon? It indicates the system is mannable but some modders may want to disable it. I tried to set `.bNeedsManned` to false. It made the function of the system available only when manned. That itself is cool, but the system is still mannable

sillysandvich:
I believe you would set `bBoostable` in the ShipSystem constructor callback.

sillysandvich:
Another thing to consider might be how to render subsystems, because subsystem positions are hardcoded and not dynamic, any user subsystems would be in a weird location.
image
And we couldn't reliably place the custom subsystems in the slots for unused subsystems because of the fact that they might have arbitrary additional UI elements, so there might need to be some special handling for this.

ranhai613:
Yeah, that should be an issue to be performed in 1.18

RenderReactorBar.cpp
```diff
@@ -36,6 +45,33 @@ HOOK_METHOD(SystemControl, RenderPowerBar, () -> void)
...
     GL_Color powerBarColour = powerBarColourOn;
 
+    //TODO: Move to constructor or initialization function later
+    if (!customWire)
+    {
+        customWire = G_->GetResources()->GetImageId("wireUI/wire_custom.png");
+    }
+    if (!customWire_connector)
+    {
+        const char* path = "wireUI/wire_custom_connector.png";
```
ranhai613, line 55: Is there any reason you define each argument value here?

sillysandvich: No there isn't, it was just for clarity while I was figuring out what the render functions did.

ranhai613:
Linux sisg confirmed👍

CustomSystems.cpp
```diff
@@ -33,7 +34,44 @@ void ParseSystemsNode(rapidxml::xml_node<char>* node)
...
+            throw std::invalid_argument(error);
+        }
+        AddSystemName(sysName);
+    }
+}
+std::vector<std::string> CustomUserSystems::systemNames;
+std::unordered_map<std::string, int> CustomUserSystems::systemIds;
+void CustomUserSystems::AddSystemName(const std::string& systemName)
+{
+    //If order of nodes can change upon reload under reasonable circumstances, we may need to sort alphabetically or something so the order is well defined
```
ranhai613, line 55: Yes the order of nodes can be changed based on memory. Anyway we should sort ids. I dont want to leave a potential bug when using multiple custom systems

arcburnergit:
Hey silly do you think you could add an optional hotkey for a custom system.
I think we'd need a new system box callback for keydown and we'd need to be able to set the default hotkey in the XML next to customSystems.
I think temporal system already has a hotkey and its added to settings so something similar would be nice.

sillysandvich:
> Hey silly do you think you could add an optional hotkey for a custom system. I think we'd need a new system box callback for keydown and we'd need to be able to set the default hotkey in the XML next to customSystems. I think temporal system already has a hotkey and its added to settings so something similar would be nice.

I will have to check what hotkey related things are exposed to lua but this sounds like a good feature. I would also have to make sure that the hotkey is part of the text box for mousing over.

CustomSystems.cpp
sillysandvich: I thought the order of nodes was just based on what order things were in the xml, which is why you can do stuff like the `<level` parsing for temporal but the requested change has been made.

ranhai613: Thanks! Yea thats a good point. I remember the order is messed up in one of my pr so i needed to fix it. I may have mistaken it for something other.

sillysandvich:
> Hey silly do you think you could add an optional hotkey for a custom system. I think we'd need a new system box callback for keydown and we'd need to be able to set the default hotkey in the XML next to customSystems. I think temporal system already has a hotkey and its added to settings so something similar would be nice.

@arcburnergit I think this feature might be better as just implementing custom hotkeys in general so the user can create hotkeys for arbitrary features, would it be preferable to have a feature where hotkeys can be added through lua e.g.
```lua
Hyperspace.Settings.RegisterHotkey("my_hotkey_name", Defines.SDL.SOME_KEY, pageNumber, position)
```

Or should it be an xml feature e.g.

```xml
<customHotkeys>
    <hotkey name="my_hotkey" defaultKey="key" page="pageNumber" position="somePos"/>
</customHotkeys>
```

I'm inclined to make it a lua feature because custom hotkeys seem useless for xml-only modders but I am curious if there's a reason you suggested xml initially.

arcburnergit:
> I'm inclined to make it a lua feature because custom hotkeys seem useless for xml-only modders but I am curious if there's a reason you suggested xml initially.

I guess that since I figured that you wouldn't want to add or remove hotkeys while the game is running it would work better as XML.

Additionally with controllable artillery implemented in #454 it would make sense for XML modders to be able to set hotkeys for controllable artillery they put on ships, but im not sure how that would interact with custom hotkeys.

THETASOLA (approved):

ranhai613 (approved):

Issue #292, opened 2024-07-01
Number hotkeys to select crew drones + saving their crew position


JerseeBeep:
Allowing number hotkeys to hand control of crew drones like they already do with weapon aiming would help their use, same with saving their position with the "Save crew stations" base feature.

PR #438, opened 2025-01-25
Drone UI Improvements (Issue #292)

sillysandvich:
## Additions
The following features have been added and can be enabled optionally using new tags in `hyperspace.xml`:
- Crew drones can now be selected using the drone power hotkeys.
- Crew drones are now capable of saving and returning to saved stations like regular crew.
### ~~Signatures~~
~~The following ZHL signatures need to be added for Linux platforms~~
- ~~void DroneControl::SelectArmament(DroneControl *this, unsigned int i)~~
- ~~void ShipManager::SaveCrewPositions(ShipManager *this)~~

Added but untested
### ~~Structs~~

- ~~DroneBox in stripped headers. (This is likely the same for all platforms but it should be checked.)~~

Checked decompile, same for Linux.

Closes issue #292 when merged.


ranhai613:
Crashed when activating boarder drone to be flying to the enemy ship.
The crash occurs in this line:
```c++
G_->GetCApp()->gui->crewControl.SelectPotentialCrew(crewDrone, false);
```
It may fail to get a pointer of CrewDrone. My guess is that the DroneBox holds BoarderDrone, a member of BoarderPodDrone class (BoarderPodDrone itself is belong to SpaceDrone,) and the boarder drone isnt ready for something (you know its flying inside the boarder pod drone)

sillysandvich:
From what I can tell the bug is that under these circumstances, the "cast" done with manual pointer arithmetic leads to an invalid result. I'll look into either finding a way to ensure the cast remains valid or an addition condition to check before selecting crew.

ranhai613 (approved):
Bug fix and linux sigs confirmed. Great job!

Issue #439, opened 2025-01-25
Add the option to "Wait" when playing on an autoship and either your engines or piloting is fully damaged and breached


gabrielsubzero:
Currently if that happens and you can't buy breach sealant or use your drones you are softlocked with no way to continue the run. If a wait option was added it would be possible to gamble on ships passing by and that way be able to get the scrap/fuel needed for breach sealant.

PR #442, opened 2025-01-26
User can enable jump map wait menu (issue #439)

sillysandvich:
## Overview
This PR addresses the issue detailed in issue #439. Because the softlock conditions are relatively complex and the specific logic is only one instance of when a distress beacon might be desirable, the logic for activating the distress beacon is not hardcoded within Hyperspace in this implementation.
## Feature details
- Adds a `StarMap::ForceWaitMessage` function to the `Hyperspace` module of the Lua API.
## Changes
- Rewrote `StarMap::OnLoop` to force the wait menu when appropriate.
- Restored inlined function `bool GenericButton::Hovering()` to FTLGame headers
- Fixed incorrect return on `StarMap::NeighboringSector`
- Added signature for `MouseControl::SetValid(MouseControl *this, bool valid, bool newValid)` on all platforms (Tested on Windows).
## Required additions
- ~~`MouseControl::SetValid` signatures need to be added on Linux.~~
- Rewrite and signature changes should probably be tested on Linux platforms.
## Notes
- There might be some ownership issues with the new function if the GL_Primitive is garbage collected while still in use by StarMap. This shouldn't happen under normal circumstances but it is worth considering.
## Example
Patch this mod after Hyperspace in order to test the feature. It adds an augment called HEAVY_BOX that activates the distress menu when held.
[force-wait-message.zip](https://github.com/user-attachments/files/18549649/force-wait-message.zip)


sillysandvich:
Before merging, be aware this can also be implemented as a `StarMap::waitMessageOverride` visible only from the lua side. Let me know if this implementation is preferred.

THETASOLA:
Confirming the linux sig and rewrite.

Issue #413, opened 2024-12-26
Add a XML tag for weapons having a max amount of times being shot per fight.


gabrielsubzero:
Currently we need to use chains to get that effect, which works fine enough for conservative, single shot weapons, but if you want to have anything else it becomes much harder to do without it being exploited by either turning off the weapon or stopping the chains midway.

A proper tag for this would greatly open the design space for pre-weapons imo.

PR #444, opened 2025-01-27
added shotLimit tag

sillysandvich:
This PR addresses issue #413.
## Features
- Implements `<shotLimit>` tag
## Changes
- Added additional argument to `WEAPON_RENDERBOX` to accommodate new code for rendering shot limit
- Added `shotLimit` to `WeaponDefinition` 
- Rewrite of `WeaponSystem::ForceIncreasePower` to undo inlining of `WeaponSystem::PowerWeapon`
- Reimplemented inlined method `WarningMessage::Stop`
- Added new constructors for `TextString`
- Added several zhl sigs.
- Added priority to hook for `WeaponControl::SelectedArmament` lua callback so it wraps the new code
## New ZHL Signatures
- ~~`WeaponSystem::ForceIncreasePower`~~ Added
- ~~`WeaponSystem::DePowerWeapon`~~ Added
- ~~`WeaponSystem::PowerWeapon`~~ Added
- ~~`WeaponControl::RenderWarnings`~~ Added
- ~~`ProjectileFactory::LoadState`~~ Added
- ~~`WeaponControl::OnLoop`~~ Added
-  ~~`ProjectileFactory::ReadyToFire`~~ Added
- ~~`ProjectileFactory::QueuedShots`~~ Added

NOTE:

This could probably be implemented on the lua side if `HitShotLimit`  was changed to be a lua callback, and `WarningMessage` + other necessary structs were exposed. I'm not sure which kind of implementation is preferred so I'll leave this as-is for now.


THETASOLA:
I can't test right now, but looking at the decomp it would seem that `shotsFiredAtTarget` is not a variable that is saved (unless I missed something), so a player exploiting save/loading might be able to bypass the limit.
It shouldn't be much of an issue to save it since it is a plain integer.

THETASOLA:
In my opinion a lua integration is not strictly necessary for this, 90% of modder still relies on pure xml for their mod and there are few scenario where a lua modder would need to modify that value on the fly (if they do, they can easily do their own implementation). If it turns out that later people would really like this to be exposed to lua then a new PR can be made later.

sillysandvich:
Alright, I'll finalize the xml version then. Thanks for your feedback.

sillysandvich:
There's still some issues with the WarningMessage (It's the wrong color and doesn't vanish after a couple seconds) so I'll fix that when I get a chance to look at how WarningMessage works in Ghidra

ranhai613:
image
It's a shame that the "Shots Remaining" text is overwrapped with the next one. Im thinking of a workaround but i would leave it as it is if we cant find any idea.

sillysandvich:
All I can think of for this is to either change it to something shorter like "shots left:" or just using some icon instead of words

sillysandvich:
Adding QueuedShots sigs now

ranhai613 (dismissed):
Sigs confirmed. ill merge this pr after all translation is done

ranhai613:
nah i think its okay to update all translation when all reveiw is done.

ranhai613 (approved):

Issue #295, opened 2024-07-09
crew with assigned rooms in hyperspace.xml don't consistently start in those rooms on restart


ChronoVortex:
If you assign crew to specific rooms like so in `hyperspace.xml`:
```xml
<ships>
  <customShip name="PLAYER_SHIP_EXAMPLE">
    <crew>
      <human room="12" />
      <human room="10" />
    </crew>
  </customShip>
<ships>
```
Then assuming those rooms aren't systems they would default to starting in anyway, they will spawn in the wrong rooms when you restart the game during a run.

PR #449, opened 2025-01-28
custom crew restart bugfix

sillysandvich:
## Feature
This PR addresses the bug where crew positions defined by hyperspace are not respected on restart. (Issue #295).
## Implementation
`ShipManager::AddInitialCrew` handles adding crew to a ship and is called once within `ShipManager::Restart`.
`AddInitialCrew` adds all crew from `ShipManager::myBlueprint::customCrew`, so the crew added by Hyperspace are temporarily removed before `ShipManager::Restart` is called.
## Notes
There's a small piece of code in `ShipManager_Extend::Initialize` where a variable called `species` is declared and unused. It isn't clear if this is a typo when `i.species` is used instead of `species`, or if this code is unnecessary and should be removed.


ranhai613:
image
it seems you actually forgot to remove hs crews

ranhai613 (approved):

Issue #451, opened 2025-01-29
Fix the ship looking strange when cloaking with no crew.


gabrielsubzero:
![Image](https://github.com/user-attachments/assets/84392847-e8d8-4e3c-b0bc-72209ec8d402)

Currently if your ship doesn't have any crew in it and you cloak the system and tiles all vanish, leaking only the cloak image, hull image and floor image. It's pretty jarring to look at and, even if it's intentional, it doesn't really look like it.

There's 2 ways to go about it, either make the rooms visible or hide the floor image so the ship looks cleaner.

Ideally this should have an xml option to toggle it. Having the rooms be visible with sensor levels would be ideal as well.

PR #456, opened 2025-01-30
Cloak render fix

sillysandvich:
This PR addresses issue #451. 
## Changes
- When sensors provide visibility of the player ship, room interiors are shown even when no crew are onboard during a cloak.
- When sensors do not provide visibility of the player ship, the floor image is hidden.


THETASOLA:
No need to push on the `develop` branch any more we've tried it for 1.16.1 and weren't convinced, so we are circling back to pushing for main directly.

THETASOLA:
It would be great that to make this feature optional with xml.  Because it does impact balance in a specific scenario compared to vanilla.

ranhai613:
image
not sure why but floor image is shown but interior is hidden at the very moment when activating cloak. you can see this when pausing game then activating cloak

sillysandvich:
As far as I can tell, this is because `bCloaked` is false until time passes, so none of the adjusted rendering actually takes affect until you unpause.

sillysandvich:
could just change the condition to used `CloakingSystem::bTurnedOn` instead of `bCloaked` in that case?

ranhai613:
> could just change the condition to used `CloakingSystem::bTurnedOn` instead of `bCloaked` in that case?

Try it, then see if it still seems weird

sillysandvich:
I can't replicate the weird behavior with the current implementation using Crystal B in vanilla Hyperspace, is there something different about whatever ship this is?

ranhai613:
Fix confirmed. Great job👍

ranhai613 (approved):

Issue #437, opened 2025-01-25
Augmentation Value that enables global sector map names


arcburnergit:
It'd be nice to be able to enable the ability to view the entire sector map using an augment of some kind.
I imagine it would replace the "Unidentified Nebula Sector" with the actual sector name when hovering over an unknown sector.
It seems like this would be easier to do within Hyperspace rather than lua.

PR #457, opened 2025-01-31
Added SECTOR_SCANNER augment

sillysandvich:
## Feature
- Added `SECTOR_SCANNER` augment that allows the player to see the precise type of non-adjacent sectors.
- `<value>` tag describes how many sectors ahead the player can see.

Closes issue #437


ranhai613:
When you have the aug:
image
image

When you dont:
image
image

So in vanilla, tooltip of sector name is shown only when the sector is unvisited and not current/next sector. You should disable tooltip on current/next sector

sillysandvich:
Made requested change.

ranhai613:
Nice jop👍
I asked translators for the translation for the name and description of sector scanner augment.

ranhai613:
the linux sigs for `StarMap::GetPotentialSectorChoiceName` is missing. could you add them?

sillysandvich:
I don't have the linux binaries on my device at the moment, is there some database of signatures that HS devs use or should I just get the different binaries set up?

ranhai613:
> I don't have the linux binaries on my device at the moment, is there some database of signatures that HS devs use or should I just get the different binaries set up?

Even if you aren't on linux, you can get linux binaries
> From Mac or Windows you can fetch the linux copy using the [steam console](https://github.com/FTL-Hyperspace/FTL-Hyperspace/wiki/Finding-Signatures-with-Ghidra#opening-the-steam-console) `download_depot 212680 221002` (steam will list where the files were saved)

We once planning of such database but never implemented.

sillysandvich:
Thank you! I'll add signatures for whichever PRs of mine are missing them when I get the chance, though I can't currently test them.

sillysandvich:
I have the binaries analyzed but the calling conventions don't match what they should be, is this a problem?

ranhai613:
> I have the binaries analyzed but the calling conventions don't match what they should be, is this a problem? For now I'll just infer from the other calling conventions in the zhl files but if this becomes an issue I can fix it.

Yeah what you did is correct. Ghidra is dumb at guessing calling conventions. Just infer from other zhl stuff. Futher explanation about calling conventions for each os is written in Global.zhl (or maybe Globals.zhl, not sure)

ranhai613:
> Thank you! I'll add signatures for whichever PRs of mine are missing them when I get the chance, though I can't currently test them.

Feel free to leave linux testing to us, but its helpful if you can test on your end (it means reducing our work😆)

Im on windows and runing linux FTL on wsl2. I just installed steam on wsl then got FTL, installed HS (maybe there is clever way)

sillysandvich:
I'm familiar with WSL, I can try setting up FTL on that for testing in the future.

ranhai613 (dismissed):
Sigs confirmed on linux. I will merge this pr after finishing collecting all translation

ranhai613 (approved):

PR #463, opened 2025-02-03
fix a bug that a hovering info of an augment in the wrong page is shown

ranhai613:
havent tested on linux. pls test


THETASOLA:
Works on linux.

THETASOLA (approved):

Issue #387, opened 2024-11-01
Enemy non-artillery weapons with more shots than player ship rooms freeze the game.


Kommandant-Julk:
When an enemy fires a non-artillery with more shots than the player ship has rooms, the game freezes forever.
(i think this used to be an issue with LASER artilleries but was fixed, but not regular weapons)

The-Dumb-Dino:
funny bug tbh, but I don't see where this would ever affect people. I doubt there are ships with so few rooms or guns with so many shots in a normal environment. 
(I am not against fixing that tho

PR #464, opened 2025-02-03
Address issue #387

sillysandvich:
## Overview
This PR rewrites `CombatAI::UpdateWeapons` to fix an infinite loop bug when an enemy fires a weapon with more shots than available rooms to target
## Added signatures (Linux untested)
- `CombatAI::UpdateWeapons`
- `CombatAI::PrioritizeSystem`
- `ProjectileFactory::ReadyToFire`

## Testing

This still needs some testing to make sure everything works as it should, but the edge case is resolved and can be tested with the following mod:
[big-guns.zip](https://github.com/user-attachments/files/18644284/big-guns.zip)

Load in any enemy ship and run `LUA give()` in console to give the enemy a weapon that fires 100 shots. The expected behavior is that the weapon fires normally and kills you.


ranhai613 (approved):
Linux sigs confirmed

PR #466, opened 2025-02-03
Add Button_Extend

ranhai613:
Add an extended class for Button for future development. I havent testd on linux.


ranhai613:
need destructor of Button

ranhai613:
> Lotation?

:(

Issue #436, opened 2025-01-24
Random ship button rolls vanilla ships and custom ships separately


bamalf:
When we have vanilla ships and custom ships at the same time, while choosing random ship it will only choose vanilla ships until you switch to the page with custom ships. This will choose among only custom ones, but not vanilla ones.

PR #468, opened 2025-02-04
Random Ship Bugfix

sillysandvich:
## Feature
This PR fixes a bug with the random ship button where the menu would only choose vanilla ships when on the vanilla page, and only choose customs ships when on a custom page.

Closes issue #436.


ranhai613:
image

Clicking those right/left button cannot move between vanilla and custom pages. Is it possible to fix this?

sillysandvich:
The buttons now switch between vanilla and custom ships properly.

ranhai613 (approved):
Great job!

PR #285, opened 2024-05-05
New command "SWITCH [SHIPBP_NAME]" allowing the replacement of current ship with the specified blueprint

THETASOLA:
The original intention was to having this method exposed, but during testing it has been discovered that plainly switching ship (in the hangars or with the command) causes memory leak, and lua would allow for too much liberty on spamming a memory leak intensive command.

Until the cause of those leak is found (specific to HS) the method will have to remain simply unexposed in a command.

This method required the use of two new sigs (not all OS sigs have been taken, will need to add them):
- CommandGUI::LinkShip (window/AMD64)
- ScoreKeeper::SetShipBlueprint (window)

PR #471, opened 2025-02-05
Ship Switch

THETASOLA:
This is another attempt at [this feature](https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/285) starting from scratch with a cleaner method.

New command `SWITCH [shipid]` `SWITCHALL [shipid]` which switch the ship blueprint in a run
Also Hyperspace.SwitchShip(SHIPID) for lua usage

Requires less sig in general than the previous solution, no noticeable memory leak, no crash.

Untested on linux, sigs uncollected on linux.


THETASOLA:
Maybe the lua functions could be attached to WorldManager, since it is reliant on it (also would be fancier).

There is a crash case if you try to switch while ShipManager(1) is on screen, the condition is made to not allow that to avoid it, but the condition can be improved because it is very restrictive for now (blocks even if there used to be a ship on the beacon).

Maybe there could be a way to allow for switch while another ship exist by looking at the crash stack trace.

ranhai613 (approved):

PR #473, opened 2025-02-07
Oxygen mechanics now function without system

THETASOLA:
## Overview

Alters the game's native implementation of oxygen mechanics, allowing rooms to have oxygen even on ships without a system.

## Implementation Details

The implementation is covered in more detail in the comments within `OxygenWithoutSystem.cpp`.
By ensuring all ShipManager instances have an oxygenSystem even if a system is not installed and bypassing certain checks within the code, OxygenSystem methods can be called and oxygen mechanics can function. Hooks on other ShipManager functions ensure that the oxygen system is not treated as installed except when necessary.


sillysandvich:
I'm not sure if this is relevant, but I think the return of `IsSystemHacked` should be `int` rather than `bool` for both cases?
image
image

sillysandvich:
This resolves the issues in my release build:
image

PR #474, opened 2025-02-07
Custom Hotkey draft

sillysandvich:
## Feature
This PR enables the user to register custom hotkeys from lua, as well as expose some utility functions for checking hotkey 
values.
(Alternate implementation of #479)
Addresses the following request when completed: https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/433#issuecomment-2642586453
## TODO:
- The feature could also be implemented through xml parsing if that is preferable.
- `Hyperspace.Settings` already refers to a singleton instance of `SettingValues`, so it may be a good idea to rename this class.
    - The `SettingValues` instance is actually a static member of the `Settings` class so merging them might actually be the most appropriate
- Add documentation to lua wiki once names are finalized.
- ~~Possibly throw a proper lua error rather than using `ErrorMessage` if duplicate registration is attempted?~~ Done
- Possibly add saving support? This may be simpler if this is an xml only feature.
## Usage
The following method can be used to register new hotkeys through lua:
```lua
local keyName = "newHotkey" --New hotkey name
local defaultValue = Defines.SDL.KEY_SPACE --Default value for the hotkey
local page = 3; --An integer between 0-3 representing the 0-indexed index of the page that this hotkey is set from.
local position = -1 --An optional argument indicating where on the page the key will be placed. -1 indices that the key will be after all other keys currently on the page.
Hyperspace.Settings.RegisterHotkey(keyName, defaultValue, page, position)
```


sillysandvich:
Unless a lua implementation is highly preferred for some reason, xml is probably simpler on account of hotkeys being fixed on game load.

PR #479, opened 2025-02-08
Custom Hotkey draft (xml version)

sillysandvich:
## Feature
Alternate implementation of #474 using xml instead of lua.
## TODOS
- Change error handling to include line numbers in the xml.
- Possibly find cleaner implementation of `CustomHotkeyManager::StringToKey`
- Saving

### DIFF
--- a/Mod Files/data/hyperspace.xml
+++ b/Mod Files/data/hyperspace.xml
@@ -1442,6 +1442,17 @@ Syntax ->
 	</system>
 </customSystems>
 
+<!--Custom hotkeys-->
+<customHotkeys>
+	<!--
+		<hotkey name="my_hotkey" page="3" default="SPACE" position="-1"/>
+		name: The name of the hotkey. Requires a corresponding <text> node.
+		page: A number 0-3 indicating the page on which the hotkey is placed in the options menu.
+		default: An optional default for the hotkey. 
+		position: An optional index for the hotkey on the given page. Defaults to -1, indicating the end of the page. 
+	-->
+</customHotkeys>
+
 <!-- Stat boosts
 
 NOTE:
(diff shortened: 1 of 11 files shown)


sillysandvich:
Should the static `Settings` C++ functions be merged in to the lua `Settings` table? This currently refers to a singleton of `SettingValues` which is a static member of the `Settings` class in C++.

Mod Files/data/hyperspace.xml
```diff
@@ -1459,6 +1459,17 @@ Syntax ->
 	-->
 </customSystems>
 
+<!--Custom hotkeys-->
+<customHotkeys>
+	<!--
+		<hotkey name="my_hotkey" page="3" default="SPACE" position="-1"/>
+		name: The name of the hotkey. Requires a corresponding <text> node.
+		page: A number 0-3 indicating the page on which the hotkey is placed in the options menu.
+		default: An optional default for the hotkey.
```
ranhai613, line 1468: This comment is unkind to modders who dont dive into HS code thus dont know the hotkey names.
I suggest to add a link to [SDLKey Table](https://github.com/FTL-Hyperspace/FTL-Hyperspace/wiki/Lua-Defines-module#sdl-keys) in HS wiki and tell them to use the names defined there.
Therefore you will need to add prefix `SDLK_` to each matching string in `CustomHotkeyManager::StringToKey`

ranhai613:
> Should the static `Settings` C++ functions be merged in to the lua `Settings` table? This currently refers to a singleton of `SettingValues` which is a static member of the `Settings` class in C++.

Yea I think its okay

ranhai613:
> > Should the static `Settings` C++ functions be merged in to the lua `Settings` table? This currently refers to a singleton of `SettingValues` which is a static member of the `Settings` class in C++.
> 
> Yea I think its okay

Turned out it would not be a good idea. I found current name `Settings` in hyperspace.i conflicts between the `SettingValue` instance and `Settings` class in c++. We could probably avoid this conflict by tweaking but I prefer the simple way: rename `Settings` class to anothername

ranhai613:
Waiting for the way to add more hotkey pages in settings

PR #487, opened 2025-02-16
System Exclusivity

### LINKED ISSUE #482: Ability to make custom systems mutually exclusive with other systems
For example an alternative to the cloaking system that replaces the cloaking system if you buy it. If this were an option it would also need to be mutually exclusive with other systems that are mutually exclusive with cloaking.

It would be also be useful for creating alternatives to medbay and clonebay, or to shields ect.

This may rely on the ability to remove systems being implemented.

sillysandvich:
## Features
This PR adds an `exclusivityGroup` tag to `hyperspace.xml` such that the user can define mutually exclusive systems.
Additionally, `<dualMedical>` can be used in the `hyperspace.xml` to prevent medbay and clonebay from excluding each other
Closes #482 when completed.
## ZHL Changes
Adds signatures for the following functions:
- `ShipSystem::RawDecreasePower`
- `ShipObject::RemoveEquipment`
- `SystemStoreBox::GetConfirmText`

Fixes return for `DroneSystem::RemoveDrone` (was `void`, should be `Drone*`)
Fixes incorrect match for `CrewMember::SetCurrentSystem` on Windows (Previously matched `CrystalAlien::LoadState` 
## TODO:
- Systems are not designed to be removed in the base game so `ShipManager::RemoveSystem` needs to be extensively tested in order to ensure it does not cause crashes. Possible issues include dangling pointers after cleanup and save/load corruption. I have done a fair bit of testing already at this point but verification is probably needed before merging.
- ~~Saving and loading needs to be tested and debugged as well~~ Done
- System replacement needs to be tested with native and custom systems.
- ~~A warning should be added for the store, similar to how the store warns you when installing a new medical system on the ship with the old one still present.~~ Added


THETASOLA:
`ShipManager::RemoveSystem` would be a true game changer for modder.
You are probably already planning on this but having an xml tag for event and a lua exposure for it would be cherry on top.

Also quick warning about `SystemControl::CreateSystemBoxes` it seems to be faintly leaking memory, meaning that after ~300 call you might encounter a SEGFAULT about a free on a wrong size, or a dangling pointer. Windows seems fairly immune but Linux encounter it quite quickly. 

The odds of getting this is quite random (given the nature of memory leaks), that is also why currently the ship switch is not exposed to lua so that modders can't abuse this and crash by accident people's game.

Best way to test this would be to have a lua script that runs ON_TICK calling `AddSystem` and `RemoveSystem` if after 10k tries you do not crash then it should be safe to use.

From our testing the SEGFAULT often happen on the weapon and door SysBox.

sillysandvich:
> `ShipManager::RemoveSystem` would be a true game changer for modder. You are probably already planning on this but having an xml tag for event and a lua exposure for it would be cherry on top.

I can add an xml tag for events after finishing up some of the small bugs with the method, if that isn't redundant to using event hooks with the Lua method.

> Also quick warning about `SystemControl::CreateSystemBoxes` it seems to be faintly leaking memory, meaning that after ~300 call you might encounter a SEGFAULT about a free on a wrong size, or a dangling pointer. Windows seems fairly immune but Linux encounter it quite quickly.

I don't think there's any good way to avoid using `CreateSystemBoxes` because otherwise the `SystemBoxes` will have dangling pointers to the removed systems as well as an outdated UI. We could experiment with setting `SystemControl::addedSystem` to `true` so that `SystemControl::OnLoop` calls `CreateSystemBoxes` but this way seems safer to avoid any dangling pointers. It might be better to fix the issue with `CreateSystemBoxes` on linux and leave things as they are.

I'm not entirely sure on how to handle artillery systems with this. I'm thinking that internal Hyperspace usage should remove all artillery systems, and that if the method is important to lua modders they can have a `ShipManager::RemoveArtillery` that takes an index or a weapon name or system pointer. @THETASOLA I'm not sure what use cases people would actually have for removing systems, if you could provide some of them I can figure out the best way to add this to the Lua API.

THETASOLA:
> Also of note is the fact that entering combat without an engine system causes a crash, but this does not seem to be a new bug and can probably be marked as a separate issue.

Yes it is a separate issue, FTL does not really appreciate engineless ship, until we fix it (or not since it is an essential system of a ship, like piloting) I would recommend that we comment out the ability to remove the engine.

sillysandvich:
In case we want to add engineless ships in the future, the crash is caused by `WorldManager::OnLoop` calling `ShipSystem::GetPowerCap` on a null engine system.

THETASOLA:
> Before marking this ready for review I would like to resolve some bugs with crew manning and repair edge cases, because crew->SetCurrentSystem(nullptr) doesn't seem to work for this. I can look into other approaches but if anybody knows offhand of the proper way to do this that would be helpful.

For that case, I believe you will have to play with CrewAI CrewTask, you can give them a new priority task to do nothing to override manning/repair task

The best example out there is the HOOK for CrewAI::UpdateDrones, it should give you the necessary task ID's.

sillysandvich:
Turns out the signature for `SetCurrentSystem` was too short and was matching `CrystalAlien::LoadState` instead, crew edge cases should be handled now.

THETASOLA:
> Oh I definitely think that removing the artillery system should remove all artillery on the ship, I am not sure there will be a big crowd for individual artillery removing. I will ask the community and come back to you on this.

No one expressed interest of doing something that would require individual artillery removing so don't bother with this. It always can be added later down the line if a new feature would make this an interesting prospect.

> Adds function quickTest(systemId, repetitions) which removes and adds systemId repetitions times. This seems to work fine for 100 repetitions but freezes on 1000 repetitions so there might need to be some fixes made before adding proper lua exposure.

This will definitely need to be fixed because the event tag are likely affected and in our working environment the memory we use is relatively tame compared to player with 20-30 mods that might encounter this error quicker.

THETASOLA:
Well we have a bit of tacit knowledge on this, but it is criminally low compared to the time we spent investigating it.

So the issue is on the freeing of SystemBox, considering that the crash happen often on the weapon/door system (both systems have substantially big primitive attributed to them, autofire button and door control buttons)
We suspect that the free call isn't actually calling the right destructor (calling the SystemBox one instead of the polymorphed one by the child).
The virtual table of SystemBox `VTable_SystemBox` has both a `*Free` and a `*destroy`, we suspect that maybe it is the destroy that is called instead of the Free, and since childs of the SystemBox do not contain a destroy function then probably only the base one is called, leading to leaks.

No other virtual table contain a `*destroy`.

This is probably a Vanilla issue, since `SystemControl::CreateSystemBoxes` is rarely called, (one time on load, one time every system bought; and people often quit the game at the end of a run) it was never encountered/reported before.

THETASOLA:
Looking at the decomp once again, this is definitely the destroy issue, vanilla code calls `_vptr_SystemBox[2]` (*destroy) instead of `_vptr_SystemBox[1]` (*Free)
I am unsure if we can easily hook it to call the virtual *Free.

sillysandvich:
If I implement a way to disable system exclusion for medbay and clonebay, how should that be handled in the xml? It could just be a toggle like many of the other features, but it could also just be on by default and have the basegame mechanic enforced by an `exclusivityGroup` tag in the `hyperspace.xml` like so:
```xml
<exclusivityGroup>
	<clonebay/>
	<medbay/>
</exclusivityGroup>
```
Though I suppose this would break backwards compatibility with outdated versions of the hyperspace xml file.

THETASOLA:
My guess would be a simple separate xml tag like `<removeMedCloneExclusivity enabled="true">`. People may want to remove all exclusivity group and this would basically ensure it (if no <exclusivityGroup> is used).
You can put a comment saying that if the player wants to touch the medbay/clonebay exclusivity to for example add a third system then this option should be put to true.

arcburnergit:
would it be possible to disable exclusivity on specific ships, especially for medbay/clonebay?
I feel like this would be the best way to handle it becuase disabling/enabling globally would require ships to stick to one setup or the other but not both.

sillysandvich:
> would it be possible to disable exclusivity on specific ships, especially for medbay/clonebay? I feel like this would be the best way to handle it becuase disabling/enabling globally would require ships to stick to one setup or the other but not both.

I think the best way to handle this is to add an `<exclusivityOverride>` tag or something within the `<customShip>` tag so that mutually exclusive systems can be specifically defined for a ship. I will implement this when I have the chance to.

PR #488, opened 2025-02-19
lowercase command parity for sys all

sillysandvich:
When adding specific systems, `sys systemName` and `SYS systemName` were both functional, but `sys all` does not work unless capitalized.


ranhai613:
Also it would be nice to update SYS ALL command to enable to add custom systems as well

ranhai613 (approved):

PR #490, opened 2025-02-20
SystemBox vTable fix & Ship Switch improvement

THETASOLA:
SystemBox virtual table had a shift of one index causing the wrong destructor method to be called resulting in memory leaks.

This change fixed the main issue of Ship Switch as such we did another optimisation pass on the it for the occasion.


THETASOLA:
I am an utter failure.

PR #495, opened 2025-02-22
add priorityEventsOverrideNebula for sector generation

### LINKED ISSUE #494: priority events failing to generate in nebula sectors
Nebula sectors have always been janky. The game seems to generate the layout of the nebula before placing any events, which means that events defined with the `priorityEvent` tag in `sector_data.xml` which would normally be guaranteed to spawn may not in a nebula sector. Some new tags which allow for more precise control over the generation of a nebula, or a `nebulaOverride` tag for `priorityEvent` would help significantly.

This issue isn't marked as a bug since it's possible that existing mods depend on the current behavior. Changing the way `priorityEvent` itself works would break backwards compatibility.

ChronoVortex:
Closes #494. There's a couple things to do before this is ready:
- If possible, solve random generic `NEBULA` events spawning.
- Test *extensively,* nebula generation is extremely finicky and there may be edge-cases this doesn't address. We also need to make absolutely certain this doesn't break the save file.


ChronoVortex:
Couldn't figure out why `NEBULA`s were spawning without being in the sector data, there didn't seem to be any obvious reference in the decomp. Opening this for review as-is.

THETASOLA (approved):

PR #497, opened 2025-02-22
Fix for issue 491

### LINKED ISSUE #491: Targetable artillery keeps shooting after a fight when auto fire is on
A bug related to #454 

You can see a targetable artillery continue to fire after a fight when auto fire is on. Since the crew manning on weapon system can gain xp from artillery's firing, this bug leads a safe and easy xp farming glitch.

THETASOLA:
fix #491 

Clears projectile & targeting if the artillery target is considered destroyed.

### DIFF
--- a/ArtillerySystem.cpp
+++ b/ArtillerySystem.cpp
@@ -122,6 +122,13 @@ void ArtillerySystem::OnLoop_HS_ManualTarget()
     int iHacked = bUnderAttack ? iHackEffect : 0;
     projectileFactory->SetHacked(iHacked);
     projectileFactory->isArtillery = true;
+
+    // Clear aiming if the target is destroyed
+    if (G_->GetShipManager(1) && G_->GetShipManager(1)->bDestroyed)
+    {
+        projectileFactory->ClearAiming();
+        projectileFactory->ClearProjectiles();
+    }
     /*
     if (projectileFactory->ReadyToFire()) 
     {


sillysandvich:
In case this doesn't handle crew deaths, it might be better to change the condition to `!target.hostile` or something like that.

THETASOLA:
That is a fair point.

sillysandvich:
Also a bomb artillery will be able to target the player ship, so I don't know if that counts as a skill farming exploit or just a regular game mechanic to be preserved.

THETASOLA:
Well to be frank, there are many skill farming exploit, what matters is that they are long and boring to disincentivise regular player to go for them. I would say waiting for an artillery to charge and shot 50 time counts as one, so I don't think it needs fixing.

ranhai613:
> In case this doesn't handle crew deaths, it might be better to change the condition to `!target.hostile` or something like that.

`CommandGui.upgradeButton.bActive` would be a good flag because upgrade button is disabled while fighting.

> Also a bomb artillery will be able to target the player ship, so I don't know if that counts as a skill farming exploit or just a regular game mechanic to be preserved.

So this issue will be solve at the same time.

sillysandvich:
I changed the conditions so it only clears targeting when the target is the enemy ship. Since the player ship can be targeted with bomb weapons at any time this behavior should probably be preserved with artillery weapons. If it's a balance issue the modder can just use some kind of ammo constraint.

ranhai613:
Done by #504

Issue #498, opened 2025-02-22
SECTOR command's interaction with priority events breaks saves


ChronoVortex:
The `StarMap::GenerateEvents` hook in `CustomEvents.cpp` makes a call to `CustomEventsParser::GetCurrentCustomSector` to count priority events. If you use the `SECTOR` command to enter a new sector, then save and load, that `CustomEventsParser::GetCurrentCustomSector` will return the data for the previous sector instead of the current one, which potentially gives the wrong count and leads to an index out of bounds error.

ranhai613:
![Image](https://github.com/user-attachments/assets/3e1944fb-e8c4-4b1a-98cd-467c0762a310)
Here is the detail

PR #506, opened 2025-02-24
Save/Load fixes

sillysandvich:
### `CustomEvents.cpp`
This PR should fix the issue in #498 by ensuring the correct sector is generated on load even when changed via console. I haven't had the chance to test the exact issue with `priorityEvents` but this fixes some unrelated strangeness with the SECTOR command and corrects the return of `CustomEventsParser::GetCurrentCustomSector`.

### `CustomStore.cpp`
When saving and loading with a custom store open, the game would crash because of a nullptr value of `shopper` in the `SystemStoreBox::constructor` hook. This fixes that issue by linking the player ship.

### DIFF
--- a/CustomEvents.cpp
+++ b/CustomEvents.cpp
@@ -5124,7 +5124,9 @@ HOOK_METHOD(StarMap, SaveGame, (int file) -> void)
 HOOK_METHOD(StarMap, LoadGame, (int file) -> Location*)
 {
     LOG_HOOK("HOOK_METHOD -> StarMap::LoadGame -> Begin (CustomEvents.cpp)\n")
-    sectorChange = FileHelper::readString(file);
+    std::string sectorName = FileHelper::readString(file);
+    sectorChange = sectorName;
+    forceSectorChoice = sectorName;
     Location* ret = super(file);
     sectorChange = "";
     return ret;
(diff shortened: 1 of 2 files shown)


ranhai613:
> When saving and loading with a custom store open, the game would crash because of a nullptr value of shopper in the SystemStoreBox::constructor hook. This fixes that issue by linking the player ship.

FYI this crash is caused by [This commit](https://github.com/FTL-Hyperspace/FTL-Hyperspace/commit/595e3ea0a52920d2648d71abd62c7168a79a3358) due to lack of nullptr check and fixed by [this](https://github.com/FTL-Hyperspace/FTL-Hyperspace/commit/5e7f95fe0ea2cd724dd7b6db73dbb48573c6a5d2). Anyway passing player ship sounds makes sense.

sillysandvich:
> Glad to hear you also have caught the crash too. I really hope we can share info more robustly, but I cannot compel you to join Multiverse discord server so...😔

If that'll make collaboration easier in the future we can communicate there.

ranhai613:
> > Glad to hear you also have caught the crash too. I really hope we can share info more robustly, but I cannot compel you to join Multiverse discord server so...😔
> 
> If that'll make collaboration easier in the future we can communicate there.

Im glad you are up to it! Im pretty sure we can collaborate way more smoothly if you communicate there. Because we usually discuss HS development and communicate with other contributors there.

If you are okay, please [join Multiverse server](https://discord.gg/ftlmultiverse) and send message in hyperspace develop to notify us

ranhai613:
Back to `SECTOR` command fix, this approach turned out not working properly.

Before saving
image

After loading
image

PR #525, opened 2025-03-03
Disable low o2 warning sound with no o2 ships

ranhai613:
HS tweaks low o2 warning not to be displayed with no o2 system ships, but its not perfect because you can still hear the warning sound at the beginning of a run.

This pr disables the low o2 warning sound when the ship doesnt have o2 system.


Misc.cpp
```diff
@@ -76,29 +76,26 @@ static bool displayWarning = true;
 HOOK_METHOD(ShipStatus, RenderEvadeOxygen, (bool unk) -> void)
 {
     LOG_HOOK("HOOK_METHOD -> ShipStatus::RenderEvadeOxygen -> Begin (Misc.cpp)\n")
-    if (!ship->HasSystem(ShipSystem::NameToSystemId("oxygen")) && ship->GetOxygenPercentage() <= 24)
+    if (!ship->HasSystem(SYS_OXYGEN) && ship->GetOxygenPercentage() <= 24)
     {
         displayWarning = false;
```
sillysandvich, line 81: The warning message should also be stopped here to account for edge cases where the oxygen system is removed via features introduced under PR #487. `oxygenMessage->Stop();` should be fine for this.

sillysandvich (changes requested):
With these changes the feature works on Windows. I have not tested on the other binaries but I did check for inlining and `WarningMessage::Start` is not inlined in either of the linux binaries, so there should not be a problem there.

ranhai613:
I also fixed other argments' names in ShipStatus

sillysandvich (approved):
Everything seems good to me.

THETASOLA:
PAIN
image

THETASOLA (approved):

PR #527, opened 2025-03-03
Prevent the scrolling tooltip from overlapping the point of the cursor

ranhai613:
Fixed an issue where the scrolling tooltip obstructs vision of the crew itself, which makes it difficult to aim crewkill weapons on top of the crew

Before, 
image

After:
image


sillysandvich:
While this issue is relevant, it may be a good idea to use scrolling to prevent this from happening even for non-scrolling tooltips.
For example:
image

THETASOLA (approved):

PR #545, opened 2025-03-14
New tag to make store no longer dissapear due to hazard

### LINKED ISSUE #540: Store disappears once exited with dangerous hazards present
If the player enters a beacon where there's a threatening hazard (i.e. asteroids, solar flare, pulsar, or ASB) and a store spawns there due to various events but the hazard doesn't get cleared at any point (i.e. the <removeHazards/> tag isn't used), the store becomes inaccessible after the player exits it due to the hazard's presence.

THETASOLA:
Adds the permanentStore tag that ensure the store button remains during an hazard.

fix #540

### DIFF
--- a/CustomOptions.h
+++ b/CustomOptions.h
@@ -90,6 +90,8 @@ public:
     Setting<bool> cloakRenderFix;
 
     Setting<bool> enhancedCloneUI;
+
+    Setting<bool> permanentStore;
     
     Defaults defaults;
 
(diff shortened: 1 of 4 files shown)


THETASOLA:
Unreliable behavior when fighting ship in an hazard that contain a store.

Also maybe make it not a global variable and a store tag parameter? (I'm unsure, it seems extremely situational)

PR #548, opened 2025-03-14
added TELEPORT_RECALL augment

sillysandvich:
This PR implements the scrapped vanilla augment "Personal Teleporters" which recalls crew from a destroyed/escaping ship automatically.


sillysandvich:
Before merging I am curious if this should also recall player crew when the player leaves a battle with boarders on the enemy ship or not.

THETASOLA:
> Before merging I am curious if this should also recall player crew when the player leaves a battle with boarders on the enemy ship or not.

Yes I thing that should be a case handled by the augment.

PR #566, opened 2025-03-16
Clarify how "all" works for systemList and systemPowerDependency

Choosechee:
The "all" tag for the systemList and systemPowerDependency tags used to define stat boosts is not explained. It is not apparent from its name exactly how it works, as it only causes vanilla systems (and not reactor) to be added to the lists. Without this clarification, many would think that it included the Temporal system and custom systems, but it doesn't. I have added clarifications on how it works to fix this.


sillysandvich:
I personally think that this should be closed and reopened as an hs bug issue?

THETASOLA:
As sillysandvich stated it is better to fix the issue than to document it and "live with it".
https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/568

PR #568, opened 2025-03-17
fix how "all" works for systemList and systemPowerDependency

sillysandvich:
This should address this issue mentioned in #566, where "all" does not apply to temporal/custom systems.

### DIFF
--- a/StatBoost.cpp
+++ b/StatBoost.cpp
@@ -4,6 +4,7 @@
 #include "CrewMember_Extend.h"
 #include "Room_Extend.h"
 #include "StatBoost.h"
+#include "CustomSystems.h"
 #include "CustomCrew.h"
 #include "CustomAugments.h"
 #include "TemporalSystem.h"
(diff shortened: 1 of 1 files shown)


Choosechee:
Is there not a backwards compatibility issue with fixing this? Couldn't someone have realized how it works and made their stat boost work with the assumption that it only affects vanilla systems? I do think "all" should actually mean all systems, but maybe this should wait until 2.0? Unless that's the plan already.

THETASOLA:
> Is there not a backwards compatibility issue with fixing this? Couldn't someone have realized how it works and made their stat boost work with the assumption that it only affects vanilla systems? I do think "all" should actually mean all systems, but maybe this should wait until 2.0? Unless that's the plan already.

There will be no backward compatibility tag with this fix, the documentation state that all systems are included, the fact that custom HS system were not in it, is within the realm of a HS bug. For HS bugs it is the responsibility of the modder to not exploit them as those would inevitably get fixed in the future.

ranhai613 (approved):

PR #570, opened 2025-03-17
Controllable artillery bugfixes

sillysandvich:
## Bugfixes
This PR addresses some bugs with aimable artilleries:
- Targeting button now renders underneath status icons when position is dynamic (fire/sabotage/etc).
- BURST artilleries now function properly when targeted manually.
## ZHL Changes
- Added new `ArtilleryBox::_HS_GetHeightModifier` method.
- Modify vtable of `ArtilleryBox` to use this function. This should be tested on all platforms to ensure proper function.


ranhai613 (approved):
The fix confirmed on Linux

PR #573, opened 2025-03-19
Fix bug with engineless player ships

sillysandvich:
This needs to be tested on release builds for all platforms as it relies on a comparison of `this` to `nullptr`. The volatile variable should prevent this from being optimized out and does so successfully on windows builds but should be checked just in case.


sillysandvich:
Functions properly on release when built with -fno-delete-null-pointer-checks.

ranhai613 (approved):

PR #576, opened 2025-03-20
Add aug/cargo slot augment

ranhai613:
Adds augment `AUGMENT_SLOT` which increases aug slots and `CARGO_SLOT` which increases cargo slots.

### DIFF
--- a/CustomEquipment.cpp
+++ b/CustomEquipment.cpp
@@ -176,7 +176,9 @@ HOOK_METHOD_PRIORITY(AugmentStoreBox, CanHold, 9999, () -> bool)
     LOG_HOOK("HOOK_METHOD_PRIORITY -> AugmentStoreBox::CanHold -> Begin (CustomEquipment.cpp)\n")
     // Rewrite to get over vanilla hard-coded augment slots limit.
 
-    if (!shopper || CustomShipSelect::GetInstance()->GetDefinition(shopper->myBlueprint.blueprintName).augSlots <= shopper->GetAugmentationCount()) return false; // in vanilla, augSlots is hard-coded to 3.
+    int augMax = CustomShipSelect::GetInstance()->GetDefinition(shopper->myBlueprint.blueprintName).augSlots;
+    if (shopper->HasAugmentation("AUGMENT_SLOT")) augMax += static_cast<int>(shopper->GetAugmentationValue("AUGMENT_SLOT"));
+    if (!shopper || augMax <= shopper->GetAugmentationCount()) return false; // in vanilla, augMax is hard-coded to 3.
 
     bool ret = true;
     if (shopper->HasAugmentation(blueprint->name))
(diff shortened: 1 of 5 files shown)


ranhai613:
[additionalSlotsAugment.zip](https://github.com/user-attachments/files/19361162/additionalSlotsAugment.zip)
Here is a test mod, adding some augments:
- `AUGMENT_SLOT`: Adds 1 aug slot
- `AUGMENT_SLOT2`: Adds 10 aug slots
- `AUGMENT_SLOT3`: Removes 1 aug slot
- `CARGO_SLOT`: Adds 1 cargo slot
- `CARGO_SLOT2`: Adds 10 cargo slots
- `CARGO_SLOT3`: Removes 1 cargo slot

Also, vanilla fed 1 ship has starting `AUGMENT_SLOT` and `CARGO_SLOT` augment.

THETASOLA (approved):

AlexanderAllen:
@ranhai613 There's some bugs w/ either the demo or the augment.

When `AUGMENT_SLOT2` is added using `additionalSlotsAugment.zip` and another augment drops from an event, `AUGMENT_SLOT2` goes away instead of giving an option of which augment to keep/drop. Augment list then proceeds to get messed up.

ranhai613:
> When `AUGMENT_SLOT2` is added using `additionalSlotsAugment.zip` and another augment drops from an event, `AUGMENT_SLOT2` goes away instead of giving an option of which augment to keep/drop. Augment list then proceeds to get messed up.

I cant replicate it. It would be very helpful if you can elaborate how to replicate the bug.

> Also not sure when/how Vanilla fed gets these augments by default, I'm not seeing it using the latest available versions of `HS 1.19.0` with `MV 5.4.6`.

I didnt mean to use this medo with MV but vanilla HS. Please replace MV with Hyperspace.ftl then see if it now works.

AlexanderAllen:
Ahhh Vanilla HS. Gotcha, I haven't tested there, my bad. It's hard to
replicate on MV, since it doesn't happen 100% of the time, but like 30% of
the time. I think usually when an extra augment is gifted from an event. I
need to gather specific replication details. This nod/change is fire 🔥 BTW.

AlexanderAllen:
Replicated on Sylvan entry to Nexus event. Had both the AUGMENT_SLOT2 and CARGO_SLOT2 on. When killing Sylvan dropped two augments at the same time, AUGMENT_SLOT2 got removed, deleting permanently the locked augment that came with my ship, along with the extra space provided by augment slot 2.

To resolve this in game I have to find another shop, find the name of the original starting-ship (locked) augment I lost, re-add augment slot 2, then re-add the original augment. Pain the ass :(

You know I'm still using augment slot 2, bugs and all.

AlexanderAllen:
https://github.com/user-attachments/assets/437702b2-4d4c-4093-acbf-5fde496d934b

Here's a replication scenario for reference.
Pretty much happens anytime anyone in MV does a drop, but guaranteed when events drop two items, especially two augments (Sylvan Nexus entry event).
Cargo 2 and Slot 2 maybe don't play together well too.

ranhai613:
@AlexanderAllen Thank you for all your elaboration. I checked it and confirmed the bug was happened to be fixed by #631 lol. You can confirmed it by using that dll if you want. You can download it here https://github.com/FTL-Hyperspace/FTL-Hyperspace/actions/runs/14694481207/artifacts/3017692349

PR #578, opened 2025-03-21
Stop autofired charge guns from shooting more shots than they should

sillysandvich:


sillysandvich:
Realized there may be issues with weapons that have both <shots> and <projectiles>, may have to investigate handling of `targets` further.

sillysandvich:
Should be fine now as `<chargeLevels>` overrides `<shots>` so the special case should take care of autofire.

ranhai613 (approved):

PR #587, opened 2025-03-25
Expose ShipRepairDrone members and fix inheritence

sillysandvich:
This PR adds exposures for `ShipRepairDrone` members as well as exposing it as a derived class of `CombatDrone`.

### DIFF
--- a/FTLGameELF32.h
+++ b/FTLGameELF32.h
@@ -7328,8 +7328,10 @@ struct ShipRepair
 {
 };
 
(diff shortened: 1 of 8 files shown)


ranhai613 (approved):
SpaceDrone downcasting works. As for two newly exposed fields, `CachedImage repairBeam` does nothing for now because CachedImage isnt exposed. Setting value to an item of `std::vector<float> repairBeams` doesnt work. It seems the value is immediately set back.

PR #588, opened 2025-03-25
No crash on invalid medical replacement

sillysandvich:
This PR address issue #585 by preventing certain invalid system additions from causing crashes.


sillysandvich:
Now the behavior is the same as when attempting to install a system that cant be installed, but that behavior should probably also be changed so invalid installation options are greyed out.

ranhai613 (approved):

PR #590, opened 2025-03-26
More StarMap exposure

THETASOLA:


sillysandvich:
A note: `dangerZoneRadius` is used, but was likely `const` and optimized out of the binary. When `StarMap` functions are rewritten it can likely be exposed if referenced correctly in the rewrites.

ranhai613 (approved):

PR #593, opened 2025-03-28
Customizable System UI

sillysandvich:
## Features
- Ability to reorder custom and native system positions in the UI.
- Proper implementation of custom subsystems.
- Custom spacing now works for native systems.
## Compatibility
- `SystemBox_Extend::xOffset` has been depreciated in favor of `ShipSystem_Extend::xOffset`. It will still function correctly for custom systems as previously implemented, but will not interact properly with the new features in this PR.
## ZHL Changes
- `SystemControl::CreateSystemBoxes` has been refactored to accomodate static and dynamic rendering for subsystems.
- `SystemControl::OnRender` has been hooked and rewritten.
- Added signature for `SystemControl::UpdateSubSystemBox` on all platforms. (Note: There are two UpdateSubSystemBox functions, one of which calls the other. The callee is the hooked function here.)


libzhlgen/test/functions/ELF_amd64/1.6.13/SystemControl.zhl
```diff
@@ -4,6 +4,8 @@ cleanup __amd64 SystemBox* SystemControl::GetSystemBox(SystemControl *this, int
 static cleanup __amd64 SystemControl::PowerBars* SystemControl::GetPowerBars(int width, int height, int gap, bool useShieldGap);
 ".415741564989fe4155415455534883ec28488b3f":
 cleanup __amd64 void SystemControl::RenderPowerBar(SystemControl *this);
+".??54??89fc5589f55389f3??83ec50e8????????":
```
Nasa62, line 7: It's unlikely on the AMD64 one that you needed the two `??`'s just sayin.
You don't have to change it, but I found that Ghidra likes to wildcard the prefix bits when they're actually always the same for GoG & Steam. Doesn't hurt anything either, but this was a common thing I found.

sillysandvich: I'll make note of this for when/if I get the GoG binary.

ranhai613 (approved):
Linux sigs confirmed

PR #595, opened 2025-03-30
Fix multi overcapacity duplicate bug

ranhai613:
To replicate this bug in MV (make sure multipleOverCapacity is enabled):
1. Fill your cargo fully.
2. Go to event `QUEST_GATLING_BASE` and dig up Jackson Newman.
3. Before starting fighting against elite fed, swap over capacitied Gatling Base with any item in your cargo.
4. Destroy elit fed.
5. You have duplicated Gatling Base in over capacity box and your cargo

### DIFF
--- a/libzhlgen/test/functions/ELF_amd64/1.6.13/EquipmentBoxItem.zhl
+++ b/libzhlgen/test/functions/ELF_amd64/1.6.13/EquipmentBoxItem.zhl
@@ -1,0 +1,11 @@
+struct EquipmentBoxItem
+{{
+	friend bool operator==(const EquipmentBoxItem &a, const EquipmentBoxItem &b)
+	{
+		return a.pWeapon == b.pWeapon && a.pDrone == b.pDrone && a.augment == b.augment && a.pCrew == b.pCrew;
+	}
+	friend bool operator!=(const EquipmentBoxItem &a, const EquipmentBoxItem &b)
+	{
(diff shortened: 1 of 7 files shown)


libzhlgen/test/functions/ELF_amd64/1.6.13/EquipmentBoxItem.zhl
Nasa62, line 1: I wonder if we should tweak the ZHL Lua generator script to add a step at the beginning for "shared ZHL files for structs" or something so we don't need to keep having cases of 3x duplicate struct additions.
Lemme know what you think, I think I can do that, although maybe struct merging would make it harder.

Nasa62 (approved):
Seems fine

libzhlgen/test/functions/ELF_amd64/1.6.13/EquipmentBoxItem.zhl
ranhai613: Im okay with current procedure; I just need to copy and pase 2 times at most. But when we start to support more versions (for example, MacOs), having 5 or 7 or more the same structs would be annoying.

ranhai613: So I guess it would be worth trying along with MacOs support

THETASOLA (approved):

PR #601, opened 2025-03-31
Fixed a tpyo in hyperspace.i (MouseControl::selling)

BobThePotat0:
In FTLGameWin32.h on line 6367 the field is listed as `GL_Texture *selling;`.

I tried grepping for cselling through the repository, and its only occurrence is in hyperspace.i.

Edit: on wiki the field is already listed as `.selling`


Nasa62 (approved):

ranhai613 (approved):
Good catch!

PR #611, opened 2025-04-06
utility macro for bounds-checking wrappers of c arrays

sillysandvich:
This PR implements a macro for exposing C arrays to lua in a safe manner.


sillysandvich:
Relevant to #590 which contains some problematic exposures.

ranhai613 (approved):

PR #612, opened 2025-04-06
Updated Lua-Hyperspace-Module.md

BobThePotat0:
1. Cosmetic changes from `returnType :Method` to `returnType` `:Method`, as well as some additions that were missing in the docs but present in hyperspace.i.

2. I have one question that is related to this PR. (attachment related to text below it)
---
image
---
The GetCloneReadyList method of the CrewMemberFactory gets extended in hyperspace.i (see attachment above), such that the return type for that is "std::vector<CrewMember*>", however in the docs it is instead currently listed as "void" - and so my question is whether or not that's intentional and I'm missing something. 

(In this PR the "void" in the docs is replaced by the appropriate "std::vector<CrewMember*>")


THETASOLA:
> The GetCloneReadyList method of the CrewMemberFactory gets extended in hyperspace.i (see attachment above), such that the return type for that is "std::vector<CrewMember*>", however in the docs it is instead currently listed as "void" - and so my question is whether or not that's intentional and I'm missing something.

That is a mistake because the base ftl method returns void, I will fix that.

THETASOLA (approved):

PR #615, opened 2025-04-09
Put separated enum headers into one file

ranhai613:
We dont want HS repository to bloat by making a new header file every time you salvage enum from base game.
This pr incorporates such enum into global.zhl, eliminating unnecessary headers.


The-Dumb-Dino:
Add zhl stuff for Mac

PR #620, opened 2025-04-13
Allow multiple fights per beacon

sillysandvich:
## Feature
This PR fixes a bug where only ship could be loaded/fought per beacon. Previous behavior would keep the original ship when attempting to load a new one and only change the hostility of that ship.
## ZHL Changes
- The signature for `ShipManager::KillEveryone` was added on all platforms.


sillysandvich:
This could possibly be restructured to be implemented on a per-event basis if that is necessary, like with `checkCargo` and other features.

PR #655, opened 2025-05-28
Fix unsafe modification of `TriggeredEvent::eventList`

afishhh:
`TriggeredEvents.cpp` will happily modify an `unordered_map` while iterating it if Lua asks it to. This is, of course, undefined behaviour.

To keep current code working, we have to support at least creating new events while iteration is ongoing, looking at the iteration sites it looks like most of them will be fine if we don't specify whether modifications will be seen or not by the iterating loop.

The easiest way to allow this then is by switching to an ordered map which does not invalidate iterators on insertion and checking whether a map is currently being iterated when removing elements (because this *does* invalidate the removed iterator).

Here I implemented this by declaring some macros and guarding the relevant locations with them. This could be done in a cleaner way, like creating a safer map abstraction that handles this all for us, however this would introduce quite a bit more complexity so I opted not to do it right away. If an abstraction like that is desired I could take a look at implementing it later.


afishhh:
It's been pointed out in the discord that some code operates under the broken assumption that this list can just be modified whenever and this must also be worked around some other way.

PR #685, opened 2025-08-05
Teleport hacking and recall bugfixes

sillysandvich:
### Features
- `<teleporterHackFix>`, which bypasses the normal requirements for teleportation when a hack is initiated (Hacker can be cloaked, crew can be in separate rooms).
- `<energyBypassTeleportRecall>`, which causes the enemy ship to make use of its energy shield bypass under all circumstances where it would attempt to remove crew from the player ship (running away, many systems are damaged, crew at low health).
### Organizational changes
- `InfiniteTeleport.cpp` and `NeutralTeleporterFix.cpp` were merged into `TeleporterBugFixes.cpp`.
- The bugfix in `NeutralTeleporterFix.cpp` was merged into the new `ShipAI::GetTeleportCommand` rewrite.
#### New signatures
- `ShipAI::SetEvacRequest`
- `CrewAI::ReadyForTeleport`
#### Rewrites
- `ShipAI::GetTeleportCommand`
- `CompleteShip::InitiateTeleport`
#### Signature changes
- `TeleportSystem::CanRecieve` takes a boolean argument where previously it did not. (NOTE: Lua code that used this function before will now raise errors, I am unsure of how this worked previously).


ranhai613:
It would be really appreciated if you have a way to test this pr. AI behavior is vague and Im not sure when GetTeleportCommand and SetEvacRequest will be called.

TeleporterBugFixes.cpp
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        if (freeSlots > 0)
+        {
+            leavingParty = TeleportCrewShip(shipManager, teleSysRoom, false, freeSlots);
+        }
+    }
+    if (command == TELE_ARRIVE)
+    {
+        int arrivingSlots = ShipGraph::GetShipInfo(enemyShip->shipManager->iShipId)->GetNumSlots(teleSysRoom);
```
ranhai613: ShipInfo here is enemy's but `teleSysRoom` belongs to the opposite ship. You may use `teleTargetRoom` instead?

The-Dumb-Dino:
Needs Mac Signatures

ranhai613 (approved):

Issue #629, opened 2025-04-18
<instantEscape/> does not remove enemy box


bamalf:
We have an enemy box without an enemy as a result

PR #686, opened 2025-08-06
Remove ship box from UI with instant escape events

sillysandvich:
### Bugfix
The `instantEscape` tag previously did not remove the ship box and UI from the right side of the screen when called with a neutral ship. Resolves issue #629.
### ZHL Changes
New signatures on all platforms for `CommandGui::ClearLocation`.


ranhai613 (approved):
zhl sigs confirmed on Linux

PR #700, opened 2025-08-18
Fix double-delete introduced by RemoveChoice

sillysandvich:
Whenever a `LocationEvent` is created via `EventGenerator::CreateEvent` (The usual way of creating LocationEvents) it is added to the `EventGenerator`'s `trashList` for cleanup when appropriate. The HS `LocationEvent::RemoveChoice` method introduces a premature delete leading to a crash on proper cleanup.

### DIFF
--- a/FTLGameELF32.h
+++ b/FTLGameELF32.h
@@ -5752,16 +5752,15 @@ struct LocationEvent
             newChoice.text.isLiteral = true;
             newChoice.requirement = requirement;
             newChoice.hiddenReward = hiddenReward;
- 
+
             this->choices.push_back(newChoice);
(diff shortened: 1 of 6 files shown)


ranhai613 (approved):
I confirmed the fix.

The way I tested this fix:

1. Run this lua code
```lua
script.on_internal_event(Defines.InternalEvents.PRE_CREATE_CHOICEBOX, function(event)
	while event:GetChoices():size() > 1 do
		event:RemoveChoice(0)
	end
end)
```

2. Start a run, then restart the run
3. I confirmed the crash disappeared by this fix

PR #705, opened 2025-09-01
Expose some missing lua types and correct some function argument types

kokoro11:
This pull request addresses two issues that were causing problems with lua scripting:
- Several C++ functions that accept string arguments were using `std::string&` instead of `const std::string&`. This prevented them from being called from lua with string literals. These function signatures need to be corrected.
- Some class members were exposed to Lua, but their associated types were not, leading to errors when trying to access them. The necessary type definitions have now been exposed to Lua.

The following functions have been updated to use `const std::string&`:
- `ActivatedPowerDefinition::GetPowerByName`
- `ActivatedPowerDefinition::AddNamedDefinition`
- `ActivatedPowerDefinition::AssignName`
- `ActivatedPowerDefinition::AssignGroup`
- `ActivatedPowerDefinition::AssignActivateGroup`
- `ActivatedPowerDefinition::AssignReplaceGroup`
- `PowerResourceDefinition::AssignName`
- `PowerResourceDefinition::AssignGroup`
- `PowerResourceDefinition::GetByName`
- `PowerResourceDefinition::AddNamedDefinition`
- `CustomEventsParser::QueueEvent`
- `CustomShipUnlocks::CheckMultiVictoryUnlocks`

The following types have been exposed to support existing class members:
- `BoardingGoal` for `Hyperspace.CrewMember.boardingGoal`
- `vector<vector<Animation>>` for `Hyperspace.CrewMember.crewAnim.anims`
- `vector<ActivatedPowerDefinition*>` for `Hyperspace.CrewMember.extend.powerChange`
- `vector<bool>` and `vector<vector<bool>>` for `Hyperspace.CrewMember.skillsEarned`
- `vector<vector<AnimationTracker>>` for `Hyperspace.CrewMember.skillUp`

### DIFF
--- a/ShipUnlocks.h
+++ b/ShipUnlocks.h
@@ -66,7 +66,7 @@ public:
     void CheckSectorUnlocks(const std::string& currentShip, int sector);
     void CheckMultiUnlocks();
     void CheckMultiVictoryUnlocks();
-    void CheckMultiVictoryUnlocks(std::string &victoryShip, std::string &victoryType);
+    void CheckMultiVictoryUnlocks(const std::string &victoryShip, std::string &victoryType);
     void CheckBasicUnlock(const std::string& currentShip, ShipUnlock::UnlockType type);
     bool CheckVanillaUnlocks(const ShipUnlock& unlock, const std::string& currentShip);
     void UnlockAllShips();
--- a/CustomCrew.h
+++ b/CustomCrew.h
@@ -311,13 +311,13 @@ struct ActivatedPowerDefinition
         powerDefs.push_back(this);
     }
 
-    void AssignName(std::string &_name)
+    void AssignName(const std::string &_name)
     {
         name = _name;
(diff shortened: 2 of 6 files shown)


ShipUnlocks.h
```diff
@@ -66,7 +66,7 @@ class CustomShipUnlocks
     void CheckSectorUnlocks(const std::string& currentShip, int sector);
     void CheckMultiUnlocks();
     void CheckMultiVictoryUnlocks();
-    void CheckMultiVictoryUnlocks(std::string &victoryShip, std::string &victoryType);
+    void CheckMultiVictoryUnlocks(const std::string &victoryShip, std::string &victoryType);
```
ranhai613: The second arg also need to be corrected to const?

kokoro11: fixed in new commit

Issue #726, opened 2025-12-09
DRONE_COLLISION internal event is not called when an ion projectile collides with a space drone.


CommunistMountain:
Lasers/missiles colliding with space drones work fine.

PR #727, opened 2025-12-09
Drone ionization duration fix, and Normal damage vs Ion damage precendence fix

CommunistMountain:
1. Drone being hit by ion projectile takes the max of its current stun and the incoming stun, to avoid shorter stuns overwriting longer stuns
2. Projectile normal damage blowing up the drone takes precedence over ion damage stunning the drone. Makes sense when you think of the scenario where a drone in the middle of being stunned can be blown up by a normal projectile.


CommunistMountain:
The latest commit also resolves #726.

PR #728, opened 2025-12-11
Solve Reward-related bugs such as scaling sector scrap, and autoReward overwriting item_modify

CommunistMountain:
Hardcoded/Piecewise sector scrap amounts are now truly 1-for-1, instead of taking difficultyAmount and the std::max floor into account.


The-Dumb-Dino (dismissed):

CommunistMountain:
TODO: When checking if the player can click an option with some cost, only consider item_modify.

It will be ready for review once these are done

PR #735, opened 2025-12-31
Evasion percentage box appears red if it is less than 0

CommunistMountain:
Currently, it only appears red if it is exactly 0, and changes back to white if below 0. This PR makes it so it is still red with negative evasion


The-Dumb-Dino (approved):
What do I say except this works fine? That rewrite was from me even

PR #765, opened 2026-06-18
Improved build instructions for Linux

asoln:
The build instructions for Linux are outdated, so I've rewritten them. This is my first time making a pull request. Please let me know if I've made a mistake or anything like that. 

I see that the build instructions on Windows are added to the wiki. I've added the Linux instructions markdown file to the wiki folder, but I believe that it has to be added to the wiki manually? I'm certain someone more experienced will be able to answer this.


The-Dumb-Dino:
could check this with a clean linux install in the comming days

The-Dumb-Dino:
Read through the entire guide on my trainride; A thing worth mentioning is Docker Desktop. As far as I remember it was simpler to install and allows building just fine. Some users may prefer it

asoln:
From what I can see, the difference isn't meaningful. The [recommended way to install Docker Desktop on Ubuntu](https://docs.docker.com/desktop/setup/install/linux/ubuntu/) is to add the apt repository, then download the .deb file, then install. By just installing the engine we don't need to download the .deb file, we can just get everything through apt in the terminal. 

One thing I glossed over when writing is [the convenience script](https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script). It should just do most of the installation process automatically, and from what I've gathered it's fine to use for development. It also works across multiple distros, as long as we add a note at the end for RPM-based distros. Perhaps this should be the recommended way? EDIT: Recommended as in "You should double check the Docker documentation before doing anything".

The-Dumb-Dino:
one slight improvement then I can merge it; Instead of re-logging consider this:
```
dino@Android-Dev:~/FTL-Hyperspace$ sudo usermod -aG docker ${USER}
[sudo: authenticate] Passwort:          
dino@Android-Dev:~/FTL-Hyperspace$ ./buildscripts/linux-1.6.13/build-releaseonly-from-docker.sh 
permission denied while trying to connect to the docker API at unix:///var/run/docker.sock
dino@Android-Dev:~/FTL-Hyperspace$ exec su - $USER
Passwort: 
dino@Android-Dev:~$ cd FTL-Hyperspace/
dino@Android-Dev:~/FTL-Hyperspace$ ./buildscripts/linux-1.6.13/build-releaseonly-from-docker.sh 
Unable to find image 'ghcr.io/ftl-hyperspace/hs-devcontainer:v2' locally
...
```

`` exec su - $USER`` for a new terminal session

The-Dumb-Dino:
> From what I can see, the difference isn't meaningful. The [recommended way to install Docker Desktop on Ubuntu](https://docs.docker.com/desktop/setup/install/linux/ubuntu/) is to add the apt repository, then download the .deb file, then install. By just installing the engine we don't need to download the .deb file, we can just get everything through apt in the terminal.
> 
> One thing I glossed over when writing is [the convenience script](https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script). It should just do most of the installation process automatically, and from what I've gathered it's fine to use for development. It also works across multiple distros, as long as we add a note at the end for RPM-based distros. Perhaps this should be the recommended way? EDIT: Recommended as in "You should double check the Docker documentation before doing anything".

Would be a cool addition, you can even bounde that in another script that does all for you like the one that sets up macOS build tools found in ``buildscripts/ci/setup-macos.sh``, though this is optional.

The-Dumb-Dino:
since I can not test the buildscript atm I have to remove this from the 1.23.0 milestone

asoln:
I'm still having some trouble with my own VMs, but I'll send a message once I've sorted and double checked it.

asoln:
After finally getting the VMs in order, I've done a clean build on Ubuntu 26.04 and confirmed that the instructions do work.

The-Dumb-Dino (approved):

PR #775, opened 2026-08-02
Oxygen refill values fix

### LINKED ISSUE #673: Oxygen refill values are incorrect
Levels 2, 3, and 4 currently have 3x, 6x, and 9x, when they should be 4x, 7x, and 10x. The easiest way to test this is to go to 0 oxygen, pause, put oxygen to the desired level, and also power up any weapon. At the 5-second mark, the value for level 1 oxygen should be 6%, and for the other levels it should be 24%, 42%, and finally 60%.

asoln:
**Summary**
Due to a coding or graphics oversight, the oxygen refill values do not reflect what is written in the menu when hovering over the oxygen system. This PR changes the oxygen refill values to the ones displayed in-game, meaning that they now follow the pattern of increasing with 3 every extra power level the system has. (1 power gives 1 refill, 2 gives 3, 3 gives 6, etc.)

**Changes**
Adds hooks to replace `OxygenSystem::GetRefillSpeed()` and `OxygenSystem::OnLoop()`  with the correct refill multiplier calculation, as well as adding a redeclaration of `OxygenSystem::RedistributeOxygen()` necessary for the `OxygenSystem::OnLoop()` hook to function properly. 

**Testing**
Has been tested (and is working) on Fedora 44. 
I have not added any new signatures to ZHL, so I believe this PR should work fine on Windows and MacOSX. Still, I would appreciate it if someone else running these could test it.

**Related issues**
Closes #673 

Thanks to Dino and CanadaHonk on Discord for contributing code, helping with reverse-engineering / Ghidra and helping with ZHL.


asoln:
As this fix is a nerf, there is now an oxygenRefillFix option in hyperspace.xml that can be enabled should we wish to include the fix. In the future we might want to try and see if we can update the in-game graphic to accurately reflect the default values, should we wish to keep the fix optional.

ranhai613 (approved):

PR #798, opened 2026-08-08
Fix Bug reporter

The-Dumb-Dino:
- Button tooltip is translated
- Added translations
- Put more-info button back in its original place


The-Dumb-Dino:
Bug report button needs to be made exclusive to the escape menu.
Missing korean translation.

ranhai613 (approved):

Issue #786, opened 2026-08-06
Hook OnLanguageChange Functions for custom Hyperspace texts


The-Dumb-Dino:
Changing the language does not change custom texts added by hyperspace in a few menus. This also breaks some dynamically sized menus which should adapt to the string size of other languages.

The-Dumb-Dino:
I did remember correctly, I reported this before: https://github.com/FTL-Hyperspace/FTL-Hyperspace/issues/156

Well now I am a dev and know more

PR #801, opened 2026-08-10
centered seedbox for all languages in the victory screen

The-Dumb-Dino:
This makes the seedbox be centered for all langages in the victory screen. I observed it being off-center in german before, multiple times. (this->position is likely different for individual languages to center the main window)

Also simplifies some calculations which were previously done.


The-Dumb-Dino:
Not exactly the cause here but this is related to this issue:
https://github.com/FTL-Hyperspace/FTL-Hyperspace/issues/786

ranhai613 (approved):

PR #803, opened 2026-08-12
Expose Hotkey Getters

arcburnergit:
Add helper functions to SettingValue struct that call the hotkey getter functions.

```lua
    Hyperspace.Settings:GetHotkey("activate_cloak")
    Hyperspace.Settings:GetHotkeyName("activate_cloak")
```

`Settings` struct has not been exposed as `SettingValue` struct is already occupying the `Settings` namespace in lua.


ranhai613:
Its kinda ugly that those two functions look like instance methods whereas they are actually static, but we can only blame those who decided to use `Settings` without consideration.

ranhai613 (approved):

PR #816, opened 2026-08-23
Adjust credits to feature the new contributors and freed

The-Dumb-Dino:


Mod Files/data/text_misc.xml.append
ranhai613, line 1: I think hs devs name shouldnt be in credits_hyperspace_contributors_names?

ranhai613 (approved):

PR #824, opened 2026-09-11
DEFENSE DRONE FIX FIX FIX

Ataraxia-Mechanica:
In the original defense drone fix the line that determines which direction the drone should swivel towards is wrong half the time leading to drones wiggling while facing backwards:
image
image

https://github.com/user-attachments/assets/d52e3796-09b6-4b61-9100-223c379cf212

Also `SetWeaponTarget()` normalizes the angle to [-180,180] instead of [0,360] like other places, and the direction determining line happens before `PickTarget()` normalizes it back to [0,360], which have also contributed to the jank:
image

This PR fixes the directions and normalizes all angles between [0, 360].
image

### DIFF
--- a/CustomDrones.cpp
+++ b/CustomDrones.cpp
@@ -782,8 +782,8 @@ HOOK_METHOD(DefenseDrone, PickTarget, () -> void)
                 if (aimingAngle < 0.f) aimingAngle += 360.f;
                 if (aimingAngle > 360.f) aimingAngle -= 360.f;
 
-                if ((desiredAimingAngle < aimingAngle  &&  bSwivelDir && aimingAngle - desiredAimingAngle <  180.f) ||
-                    (aimingAngle <= desiredAimingAngle && !bSwivelDir && desiredAimingAngle - aimingAngle >= 180.f))
+                bool bNewSwivelDir = (aimingAngle <= desiredAimingAngle || (aimingAngle - desiredAimingAngle) >= 180.f) && (desiredAimingAngle < aimingAngle || (desiredAimingAngle - aimingAngle) < 180.f);
+                if (bNewSwivelDir != bSwivelDir) //if swiveled too much and went past, snap to angle
                 {
                     aimingAngle = desiredAimingAngle;
                 }
@@ -886,8 +886,8 @@ HOOK_METHOD(DefenseDrone, PickTarget, () -> void)
                 if (aimingAngle < 0.f) aimingAngle += 360.f;
                 if (aimingAngle > 360.f) aimingAngle -= 360.f;
 
-                if ((desiredAimingAngle < aimingAngle  &&  bSwivelDir && aimingAngle - desiredAimingAngle <  180.f) ||
-                    (aimingAngle <= desiredAimingAngle && !bSwivelDir && desiredAimingAngle - aimingAngle >= 180.f))
+                bool bNewSwivelDir = (aimingAngle <= desiredAimingAngle || (aimingAngle - desiredAimingAngle) >= 180.f) && (desiredAimingAngle < aimingAngle || (desiredAimingAngle - aimingAngle) < 180.f);
+                if (bNewSwivelDir != bSwivelDir) //if swiveled too much and went past, snap to angle
                 {
                     aimingAngle = desiredAimingAngle;
                 }
--- a/CustomDrones.cpp
+++ b/CustomDrones.cpp
@@ -8,7 +8,7 @@
 #include <cfloat>
 
 bool DefenseDroneFix::active = false;
-float DefenseDroneFix::boxRange[2] = {5000.f, 5000.f};
+float DefenseDroneFix::boxRange[2] = {150.f, 150.f};
 float DefenseDroneFix::ellipseRange[2] = {50.f, 50.f};
 
 //bool g_dronesCanTeleport = false;
--- a/CustomDrones.cpp
+++ b/CustomDrones.cpp
@@ -782,7 +782,7 @@ HOOK_METHOD(DefenseDrone, PickTarget, () -> void)
                 if (aimingAngle < 0.f) aimingAngle += 360.f;
                 if (aimingAngle > 360.f) aimingAngle -= 360.f;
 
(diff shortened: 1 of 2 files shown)


Ataraxia-Mechanica:
Anybody know if +-10000 pixels is the vanilla value? I've occasionally seen anti drone drones take an extra shot, but it's kinda rare and can't replicate it. Is it vanilla behavior, or should the bounding box be changed to something lower, or just use normal targeting box?

image

Ataraxia-Mechanica:
Scrambled drones' random spin has a bunch of extra unnecessary checks that would fail and made them stationary, I made it work

Ataraxia-Mechanica:
Removed bunch of erroneous checks in scrambled drone random rotation code that made it not work.

Ataraxia-Mechanica:
The +-10000 targeting area for combat drones isn't in vanilla, it used the same bounding box as all other types:

https://github.com/user-attachments/assets/6c383870-ba51-46c8-8018-8018f7e9b2f0

Anti drone firing time set to 50ms to make it clear when it starts shooting.

I think it's pretty unnecessary to keep the +-10000 range special case, removing it to retain parity with vanilla is better.

(might be also causing targeting inaccuracies:)
image

CommunistMountain:
Is this also related to the vanilla bug where offensive drones may sometimes travel short distances before firing again? The code is meant to make the drone pick an angle more than 90 degrees from the previous angle, but if the initial angle was close to 0 and the target angle is close to 360 the drone may move by a small angle

Ataraxia-Mechanica:
> Is this also related to the vanilla bug where offensive drones may sometimes travel short distances before firing again? The code is meant to make the drone pick an angle more than 90 degrees from the previous angle, but if the initial angle was close to 0 and the target angle is close to 360 the drone may move by a small angle

Nope, the bug this pr fixes is in defense drone fix and it's caused by flipped comparsion, that sounds like the devs forgot to check for extra conditions. Didn't even know that was the cause for combat drone rapid firing but now I think about it, it has indeed only happened on the right side of the ship where angle is close to 0 and 360.

fr-eed:
Test run on emulated Linux:

https://github.com/user-attachments/assets/fa8d0580-4f71-4aec-bb6a-cc974102cf87


Also verified on Windows and MacOS*

MacOS also had another bug related to defense drones that will be fixed in a different PR: https://github.com/FTL-Hyperspace/FTL-Hyperspace/pull/832

CustomDrones.cpp
fr-eed: A bug here. one should be .y

CustomDrones.cpp
```diff
@@ -795,30 +789,20 @@ HOOK_METHOD(DefenseDrone, PickTarget, () -> void)
                     float y0;
                     float y1;
 
-                    if (weaponTarget && weaponTarget->type == 3) // combat drone
```
fr-eed, line 798: The removed weaponTarget->type == 3 check was broken but its intent was right: in vanilla, flying drones (3) and boarding pods (6) are shot at any range (targets are only limited to the same space, in ValidTargetObject). 

This can be re-implemented like so:
```c++
if ( currentTargetType == 3 || currentTargetType == 6){
    shotAtTargetId = currentTargetId;
    return;
  }
```

CustomDrones.cpp
```diff
@@ -849,49 +833,32 @@ HOOK_METHOD(DefenseDrone, PickTarget, () -> void)
...
 
-                if ((desiredAimingAngle < aimingAngle &&  bSwivelDir && aimingAngle - desiredAimingAngle < 180.f) ||
-                    (aimingAngle < desiredAimingAngle && !bSwivelDir && desiredAimingAngle - aimingAngle < 180.f))
-                {
-                    aimingAngle = desiredAimingAngle;
-                }
-            }
+        bool bNewSwivelDir = aimingAngle > desiredAimingAngle // true if need to swivel clockwise
+                ? (aimingAngle - desiredAimingAngle) > 180.f
+                : (desiredAimingAngle - aimingAngle) < 180.f;
```
fr-eed: This can be moved into a function and reused 4 times. (L765, L778, L843, L856)

```c++
static bool swivelClockwise(float aim, float desired)
{
    return aim > desired ? (aim - desired) > 180.f : (desired - aim) < 180.f;
}

```

CustomDrones.cpp
```diff
@@ -849,49 +833,32 @@ HOOK_METHOD(DefenseDrone, PickTarget, () -> void)
...
-    else
+    else //scrambled
     {
-        if (desiredAimingAngle == aimingAngle || desiredAimingAngle < 0.f)
-        {
-            desiredAimingAngle = random32() % 360;
-            if (*(int*)(&targetLocation.x) == 0xff7fffff || *(int*)(&targetLocation.y) == 0xff7fffff || desiredAimingAngle >= 0.f)
-            {
-                float speedFactor = G_->GetCFPS()->GetSpeedFactor();
+        if (desiredAimingAngle == aimingAngle) desiredAimingAngle = random32() % 360; // if arrived at angle, plot a new random one
```
fr-eed: Should be ```(desiredAimingAngle == aimingAngle || desiredAimingAngle < 0.f)``` instead.

desiredAimingAngle starts at -1.0. with DEFENSE_SCRAMBLER the scrambled branch uses it right away, so the drone turns to -1.0 first instead of a random angle

fr-eed (changes requested):

CustomDrones.cpp
Ataraxia-Mechanica: Oops, did not catch this one. Went though it again to ensure no x and y mixing up.

CustomDrones.cpp
Ataraxia-Mechanica: Flying drones and boarding pods are shot at the same range as other targets in vanilla, according to tests by both of us. No need to add this back.

CustomDrones.cpp
Ataraxia-Mechanica: Moved into a function as requested.

CustomDrones.cpp
Ataraxia-Mechanica: Added the check as requested, checked in game that it does not turn towards -1.0 first.

fr-eed (approved):

PR #831, opened 2026-09-16
Combat drone rapid fire fix

Ataraxia-Mechanica:
In vanilla, drone angles are stored between 0 to 360 in degrees, where positive x axis direction is 0 degree, and goes up clockwise.
When a combat drone decides to chose a new position to fly to, it generates a new angle, reject if this angle and current angle's difference's absolute value is smaller than 90 (`abs(newAngle - current_angle)<90`), and regenerate until it is false.
The likely intention is for the new angle to be at least 90 degrees away from the old one so it doesn't fire very rapidly, but this fails when the old angle is near 0 degrees and the new one around 360, or vise versa:
image

This PR reimplements `CombatDrone::PickDestination()` so it generates an angle between [90, 270], then adds it to the old angle and normalizes it between [0, 360), which achieves the effect of new angle being at least 90 degrees away from the old one correctly.

Also replaces the old vanilla heading code that uses `acos` for whatever reason with cleaner one via `atan2`.

Enable with `<combatDroneRapidFireFix enabled="true"/>`, disabled by default as this changes game behavior and makes combat drones perform worse.


fr-eed:
@Ataraxia-Mechanica Could you please remove all changes to FTLGame* and libzhlgen/* files?

The CombatDrone::PickDestination signatures are already on `master` branch.

Afterwards update this branch with `master` so we could run CI checks.

fr-eed:
Tested this in-game with an automated test: 2 player + 2 enemy Combat Drone 1s `COMBAT_1` .
60 seconds without the fix and 60 seconds with.

- Off: 8/78 shots were short hops (<90 degrees) near the right side, and the next shot was 109-158 frames sooner instead of the usual 196.
- On: 0/74 shots were short hops, and the drone's new position around the enemy was at least 90 degrees from its previous one

It removes the short-gap shots, so fire rate drops from 19.5 to 17.75 shots per minute.

https://streamable.com/qbte8c

fr-eed (approved):

PR #834, opened 2026-09-19
Expose ShipManager::RemoveCrewmember(CrewMember *)

hol430:
This lets you remove crew members from a ship. I didn't expose `RemoveCrewmember(int)` because it lacks native bindings on 32-bit elf platform, and it does basically the same thing anyway. I tested this only on x64 Linux.


ranhai613:
FYI we have ditched elf32 and FTLGameElf32.h(.cpp) would be removed soon. You dont need to consider about elf32 for your work

hol430:
Ah, good to know - thanks.

I'm going to close this as not needed - this functionality is better achieved via `Crewmember::Kill()` (already exposed), with noClone=true.

PR #836, opened 2026-09-19
Shield drones show Orbit Speed in description with redesigned tooltips

CommunistMountain:


CommunistMountain:
1. Add orbit speed description to shield drones, when using redesigned drone tooltips. This is already present for defense drones, we just forgot about shield drones.
2. Use tLib variable for all instances of G_->GetTextLibrary(), for standardisation

